PR #4, opened 2024-09-25
Re-Create "Formatting Output For FluSight Submission" Vignette

AFg6K7h4fhy2:
> The re-creation of the vignette _Formatting Output For FluSight Submission_ shall be completed in Python and shall make use of `arviz`, specifically the `InferenceData` objects, which constitute the new starting place for a forecast to be produced.


AFg6K7h4fhy2:
The contents of this will be covered in #10

PR #24, opened 2024-10-17
Rename `df_daily_dates_to_epiweeks` to `df_aggregate_to_epiweekly`

AFg6K7h4fhy2:
> As written, the function name and docstring suggest that this is a helper function for adding `epiweek` and `epiyear` columns to a dataframe with a date column. It might make sense to create such a function and have the `df_aggregate_to_epiweekly` function call it, but that is out of scope for now.


dylanhmorris (approved):
LGTM

Issue #32, opened 2024-10-24
Add Remaining Location Code, Abbreviation, And Table Utilities


AFg6K7h4fhy2:
To be more specific about the scope of this issue, the PR associated with this issue adds the following functions in `recode_locations.py`

* `loc_flusight_code_to_abbr`, which _Takes the location columns of a Polars dataframe (formatted as FluSight codes for US jurisdictions) and recodes it to US jurisdictional abbreviations._
* `to_location_table_column`, which _Maps a location format string to the corresponding column name in the hubserve location table. For example, "flusight" maps to "location_code" in forecasttool's location_table._
* `location_lookup`, which _Look up rows of the FluSight location table corresponding to the entries of a given location vector and format. Retrieves the rows from location_table in the forecasttools package corresponding to a given vector of location identifiers, with possible repeats._

where `forecasttools-py` already contains `loc_abbr_to_flusight_code`, which _Takes the location columns of a Polars dataframe (formatted as US jurisdictional abbreviations) and recodes it to FluSight location codes._

AFg6K7h4fhy2:
Probably some changes needed in going from "FluSight" to "hubverse".

PR #33, opened 2024-10-24
Add Remaining Location Code, Abbreviation, And Table Utilities

AFg6K7h4fhy2:
For the scope of this PR, see the associated issue: <https://github.com/CDCgov/forecasttools-py/issues/32>.


AFg6K7h4fhy2:
This PR has been marked as ready for review.

I want to further verify that the added functions work as intended (this shouldn't take much time), so, in marking the PR ready for review, I am not making the claim that the PR is ready to be merged.

AFg6K7h4fhy2:
Once a review is provided for this PR, I will remove <https://github.com/CDCgov/forecasttools-py/pull/33/files#diff-3da0137fa7c28bf7218e9a6ff9e3474bc3db4041a9bfa8210ba845eccc02e609>. At present, I have no further comments or code to add to this PR.

forecasttools/recode_locations.py
```diff
@@ -14,31 +14,159 @@ def loc_abbr_to_flusight_code(
...
+        The format string ("abbr",
+        "flusight", or "long_name").
+
+    Returns
+    -------
+    str
+        Returns the corresponding column name
+        from the location table.
+    """
+    col_keys = {
```
damonbayer, line 112: This is a dictionary, not a list of keys, so a different name would be more appropriate.

forecasttools/recode_locations.py
```diff
@@ -14,31 +14,159 @@ def loc_abbr_to_flusight_code(
...
+        "abbr": "short_name",
+        "flusight": "location_code",
+        "long_name": "long_name",
+    }
+    col = col_keys[location_format]
```
damonbayer, line 117: This itself will error if `location_format` is not a valid key.

damonbayer: If you want to raise the custom error, you need to use `col_keys.get`

forecasttools/recode_locations.py
```diff
@@ -14,31 +14,159 @@ def loc_abbr_to_flusight_code(
...
+        from the location table.
+    """
+    col_dict = {
+        "abbr": "short_name",
+        "flusight": "location_code",
+        "long_name": "long_name",
+    }
+    col = col_dict.get(location_format)
+    if col is None:
+        raise ValueError(
```
damonbayer, line 119: use [`KeyError`](https://docs.python.org/3/library/exceptions.html#KeyError)

forecasttools/recode_locations.py
```diff
@@ -14,31 +14,159 @@ def loc_abbr_to_flusight_code(
...
+    """
+    col_dict = {
+        "abbr": "short_name",
+        "flusight": "location_code",
+        "long_name": "long_name",
+    }
+    col = col_dict.get(location_format)
+    if col is None:
+        raise KeyError(
+            f"Unknown location format {location_format}. Expected 'abbr', 'flusight', or 'long_name'."
```
damonbayer, line 120: pull the names of the valid formats from the dictionary keys?

AFg6K7h4fhy2: Not sure you are questioning here.

If the name is not a valid (supported) format, then the name will not be in the dictionary keys of `col_dict` and an error will be thrown.

damonbayer: The error message says `Expected 'abbr', 'flusight', or 'long_name'."`. Those are just the keys from `col_dict`. You should use that fact to make the error message more automated. Then, it won't be out of date if you ever add new keys to `col_dict`.

AFg6K7h4fhy2: I see. Nice observation.

dylanhmorris:
As discussed in the issue, this PR would be a good place to generalize from "flusight" to "hub" in the naming of functions and tables.

forecasttools/recode_locations.py
```diff
@@ -14,31 +14,159 @@ def loc_abbr_to_flusight_code(
...
+    """
+    # get forecasttools location table
+    loc_table = forecasttools.location_table
+    # recode location codes to location abbreviations
+    loc_recoded_df = df.with_columns(
+        pl.col(location_col).replace(
+            old=loc_table["location_code"], new=loc_table["short_name"]
+        )
+    )
+    return loc_recoded_df
```
dylanhmorris, line 89: For now it is fine to have these as functions that take and return dataframes, more polars idiomatic to define a custom expression that accomplishes this https://docs.pola.rs/api/python/stable/reference/api.html

dylanhmorris: Beyond the scope of this PR, but worth doing eventually imo

forecasttools/recode_locations.py
```diff
@@ -14,31 +14,159 @@ def loc_abbr_to_flusight_code(
...
+
+    Returns
+    -------
+    pl.DataFrame
+        Rows from location_table that match
+        the location vector, with repeats
+        possible.
+    """
+    # convert location vector to string format
+    location_vector = [str(loc) for loc in location_vector]
```
dylanhmorris, line 162: In polars, I'd cast the column when creating the data frame. Indeed, arguably in R I should've written

```R
locs <- tibble::tibble({{ join_key }} := as.character(!!location_vector)) |>
```

notebooks/test_location_recoding.py
```diff
@@ -0,0 +1,70 @@
+"""
+Testing the historical recode locations
+utilities and possibly experimenting with
+additional utilities.
+"""
```
dylanhmorris, line 5: These should become formal unit tests before the PR is merged.

AFg6K7h4fhy2:
Re: <https://github.com/CDCgov/forecasttools-py/pull/33#issuecomment-2455932923>

Sounds good. My intention with this is to retain "FluSight" in any instance where the code or object deals with the submitting to the FluSight Hub and to change instances to "hub" or "Hubverse" where the code or object deals with not-FluSight-specific hubverse patterns.

forecasttools/data.py
```diff
@@ -84,8 +84,12 @@ def make_census_dataset(
             pl.col("STATE_NAME").alias("long_name"),
         ]
     )
-    location_table = nation.vstack(jurisdictions)
-    location_table.write_csv(file_save_path)
+    location_table = nation.vstack(jurisdictions).with_columns(
+        pl.col("location_code").map_elements(
+            lambda x: f"0{x}" if len(x) == 1 else x, return_dtype=pl.Utf8
+        )
```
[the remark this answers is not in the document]
AFg6K7h4fhy2: (1) `read_parquet` is being used in `__init__`, (2) `schema_overrides` is a non-existent parameter of `read_parquet`, (3) regarding `schema` (see [here]()), we have _Specify the datatypes of the columns. The datatypes must match the datatypes in the file(s). If there are extra columns that are not in the file(s), consider also enabling `allow_missing_columns`._ with the Warning: _<span style="color: red;">This functionality is considered unstable. It may be changed at any point without it being considered a breaking change</span>._, (4) `location_code` already includes "US" and is read in as `pl.Utf8`, so stating this again with `schema = {"location_code": pl.Utf8}` doesn't seem valuable



What I intend to do is use the following in `__init__.py`:

Change

```python
# location table (from Census data)
with importlib.resources.path(
    __package__, "location_table.parquet"
) as data_path:
    location_table = pl.read_parquet(data_path)
```

to

```python
# location table (from Census data)
with importlib.resources.path(
    __package__, "location_table.parquet"
) as data_path:
    location_table = pl.read_parquet(data_path).with_columns(
        pl.col("location_code").str.zfill(2)
    )
```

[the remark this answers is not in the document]
dylanhmorris: This isn't a comment on the read from parquet, it's a comment on the code that constructs the dataset.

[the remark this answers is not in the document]
AFg6K7h4fhy2: Ah, I see. I interpreted the `read_csv()` as in "where the location table data object (parquet)" is read in. 

In light of this, what would you prefer the suggested edit to be? (I am looking at `schema` and `schema_overrides` and do not see an immediate way to do something akin to `pl.col("location_code").str.zfill(2)`).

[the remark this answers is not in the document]
dylanhmorris: The source data has zero padding, so if you just read the column as a string column it will be fine.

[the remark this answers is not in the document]
AFg6K7h4fhy2: The source data does not have 0-padding. But it needs this, at least for the FluSight Hub. By default, the `location_code` is read in as `pl.Utf8`. 1 comes up as "1". However, the FluSight submissions seem to require code "1" to actually be "01", and so on. The example FluSight submission has zero padding and Polars dataframe for `location_code` (just reading in using `read_csv`) does not pad with zeroes by default.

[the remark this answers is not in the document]
dylanhmorris: I see zero padding: https://www2.census.gov/geo/docs/reference/state.txt

Screenshot 2024-11-05 at 1 52 35 PM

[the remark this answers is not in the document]
AFg6K7h4fhy2: That's so odd. 

In testing, the following returns "1" for AL.

`pl.col("STATE").alias("location_code").cast(pl.Utf8)` 

(I've had this for a long time unchanged) .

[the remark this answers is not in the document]
AFg6K7h4fhy2: Let me check over things again.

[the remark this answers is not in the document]
dylanhmorris: This is the point of my comment. At the `pl.read_csv` stage, the column is parsed as an `int`, because it contains only (zero-padded) ints. But if you explicitly tell `pl.read_csv` to parse it as a string, you will get `"01", "02",` etc

AFg6K7h4fhy2:
A warning (in __init__) I received when running tests was:

```
DeprecationWarning: path is deprecated. Use files() instead. Refer to https://importlib-resources.readthedocs.io/en/latest/using.html#migrating-from-legacy for migration advice.
    with importlib.resources.path(
```

An example fix (which I've applied to the other warnings as well) that I want to document in this PR:

```python
with importlib.resources.path(
    __package__,
    "example_flusight_submission.parquet",
) as data_path:
    example_flusight_submission = pl.read_parquet(data_path)
```

to

```python
with importlib.resources.files(__package__).joinpath("example_flusight_submission.parquet").open("rb") as f:
    example_flusight_submission = pl.read_parquet(f)
```

AFg6K7h4fhy2:
Would appreciate another inspection here as well @dylanhmorris .

forecasttools/recode_locations.py
```diff
@@ -1,44 +1,240 @@
...
     loc_table = forecasttools.location_table
-    # recode and replaced existing loc abbrs with loc codes
+    # check if values in location_col are a
+    # subset of short_name in location table
+    location_values = set(df[location_col].to_list())
+    valid_values = set(loc_table["short_name"].to_list())
+    difference = location_values.difference(valid_values)
+    if difference:
+        raise ValueError(
+            f"Some values in {difference} (in col '{location_col}') are not valid jurisdictional codes."
```
dylanhmorris, line 65:
```suggestion
            f"The following values in '{location_col}') are not valid jurisdictional codes: {difference}."
```

forecasttools/to_hubverse.py
dylanhmorris, line 1: noting that this functionality should have tests and that `get_hubverse_target_end_dates` should probably use polars expressions rather than `DataFrame(list(map))`, but that this is out of scope for this PR

dylanhmorris (approved):
LGTM

Issue #28, opened 2024-10-21
Create Tests To Ensure Correct Dating From Arviz


AFg6K7h4fhy2:

AFg6K7h4fhy2:
Closed via #29

PR #35, opened 2024-10-28
More General Utilities With Representation Of Time In idata Objects

AFg6K7h4fhy2:
Generalize utilities to add dates and times to `az.InferenceData` objects. A demonstration of these utilities can be found in #40 .


AFg6K7h4fhy2:
#28 can be covered in this PR given the PR's scope.

AFg6K7h4fhy2:
@damonbayer Can you take a quick look?

AFg6K7h4fhy2:
I expect this PR _might_ be merged today. "Might" and not "likely" is used since I expect Dylan to have some strong comments. Given the timelines, I have still retained @damonbayer as a reviewer, should he find time and or be interested in taking a look. Please message me here or on Teams if you would no longer like this to be the case. Thank you.

notebooks/test_idata_general_time_representation.py
```diff
@@ -0,0 +1,241 @@
...
+print(option_1_idata_w_dates["observed_data"])
+print(option_1_idata_w_dates["posterior_predictive"])
+
+
+# %% OPTION 2 FOR ADDING DATES
+
+# NOTE: this option will likely not work. The idea was to find the
+# correct dimension without having to explicitly write the dim name
+# out. We know that the dim is not chain or draw, but there may
+# be other dims remaining.
```
damonbayer, line 98: By convention in pyrenew, the first dimension is time, but we probably don't want to enforce such a convention here.

damonbayer:
@AFg6K7h4fhy2 I _think_ your use of the term "group" is imprecise throughout, or I am misunderstanding something.

Per [ArviZ documentation](https://python.arviz.org/en/stable/schema/schema.html#terminology, a group is:

> Group: Dataset containing one or several variables with a conceptual link between them. Variables inside a group will generally share some dimensions too. For example, the `posterior` group contains a representation of the posterior distribution conditioned on the observations in the `observed_data` group.

I think the form of the argument used in these functions needs to be a `Dimension` (and possibly a `variable_name`.

Again, per [ArviZ documentation](https://python.arviz.org/en/stable/schema/schema.html#terminology), every `variable` has a `variable_name`, and `dimensions`:

> **Dimension**: The dimensions of an object are its named axes. A variable containing 3D data can have dimensions `[chain, draw, dim0]`, i.e., its `0th`-dimension is `chain`, its `1st`-dimension is `draw`, and so on. Every dimension present in an `InferenceData` variable must share names with a coordinate. Given that dimensions must be named, dimension and dimension name are used equivalents.

The dimension name is then the name of the coordinate array:

> **Coordinate**: A named array that labels a dimension. A coordinate named `chain` with values `[0, 1, 2, 3]` would label the `chain` dimension. Coordinate names and values can be loosely thought of as labels and tick labels along a dimension, respectively.

notebooks/test_idata_general_time_representation.py
```diff
@@ -0,0 +1,241 @@
...
+# ought to be taken into account, not currently
+# taken into account; might matter.
+
+# %% OPTION 1 FOR ADDING DATES
+
+
+def option_1_add_dates_as_coords_to_idata(
+    idata_wo_dates: az.InferenceData,
+    dim_name: str,
+    group_date_mapping: dict[str, tuple[str, timedelta]],
```
damonbayer, line 33: See https://github.com/CDCgov/forecasttools-py/pull/35#issuecomment-2448190737. I think all your uses of "group" are actually "variables"?

AFg6K7h4fhy2:
I have been using `groups` only to refer to items that would be returned by `idata._groups_all`. Examples include `observed_data` or `posterior_predictive`. I don't see an imprecision but I will look through my code again.

notebooks/test_idata_general_time_representation.py
AFg6K7h4fhy2: I feel that sections such as the following distinguish between groups and dimensions fairly clearly: 

```python
if dim_name not in idata_group.dims:
    print(
        f"Warning: Dimension '{dim_name}' not found in group '{group_name}'."
    )
    continue
```

damonbayer: But not all of the variables in a group should be given the same coordinates. How do you specify the variable you want to modify?

damonbayer:
@AFg6K7h4fhy2 It would be helpful to me if you described the difference in the approaches and what has made it difficult for you to make a decision about which to use.

dylanhmorris:
Agree with @damonbayer. I also wasn't sure what key differences to evaluate when looking over them. I have some implementation thoughts as well but I think the conceptual discussion should happen first.

notebooks/test_idata_general_time_representation.py
AFg6K7h4fhy2: > But not all of the variables in a group should be given the same coordinates. How do you specify the variable you want to modify?

I think this is a useful remark. Imagine having a separate `xarray` DataSet for weekly Rt values along with an `idata` object that has the group `posterior_predictive` with daily hospital admissions stored in a variable called `obs`. I can imagine a situation where one adds the Rt values dataset as a new group to the `idata`. However, I see now that one might have in `posterior_predictive` another variable `Rt` with different coordinates. My current approach regards to the (lower likelihood) situation where another variable of interest has been added via a new group. My approach does not suit multiple variables of interest within the same group which use different coordinates.

AFg6K7h4fhy2:
> @AFg6K7h4fhy2 It would be helpful to me if you described the difference in the approaches and what has made it difficult for you to make a decision about which to use.

Wrt decision, I've been with Option 3 (the one option I used in the test file), but I was waiting for you or DHM to comment and possibly change my opinion. I see now all options need to be more granular. The difference between the options: O1 uses a single dim name and expects all groups passed to the same dim name of interest; O2 takes no dim name and gets the first non-chain non-draw dim (fails because only assumes first non-chain non-draw dim is the only one required); O3 supports unique dim per group. All fail because they assume there is only one variable of interest (the one with the size of the single provided dim).

AFg6K7h4fhy2:
Pattern-recognizing comments from Damon:

> I'm noticing a similar theme in the two reviews I have done on forecasttools-py recently. Both have been a bit difficult to review because they contain a lot of "experimentation."
> 
> Some tips / guidelines that would be helpful to me as a reviewer and may be more efficient overall:
> 
> 1. At the top of the PR, or when requesting a review, summarize what you think the PR achieves or should achieve. If the reviewer doesn't know what you are trying to do, it is very hard to assess whether you have done it. A detailed issue can also substitute for this PR summary. As a team, we should probably be better about making high-quality issues.
2. If you are thinking about multiple ways to do something and aren't sure which to pursue, let's discuss in english and pseudocode before implementing any of them.
3. If you try multiple ways of doing something, and arrive at one you like best, leave the others out of the code. If you remain a bit uncertain or recognize there are some drawbacks, you can mention "I was also thinking about doing X, but I did Y because Z." 

Response: 

> I believe experimentation is good so long as it is removed before PR is merged.
>
> (1) I agree with you that PRs & issues should have detailed information on the task at hand. In my mind, the issue should have detailed information on the task and what success looks like re: "A Problem Well-Stated Is a Problem Half-Solved". The PR can contain quoted portions from the issue to remind reviewers who've not examined the issue what the PR deals with.
> (2) Generally agree that implementations before discussion can be wasteful. However, for the PRs, these were discussed somewhat, maybe not enough or at a sufficient quality level. Sometimes as well implementations are necessary for discussions to reach a certain level of concreteness. 
> (3) Again, generally agree that descriptions of implementation uncertainty and or drawbacks are useful but also want to flag that deferring to those with more experience and or with a deeper vision of a project is an option; those being deferred to can benefit from multiple, slightly different implementations. Deliberation on which pathway to pursue can then occur faster. Problems arise when the options are majorly different. Problems also arise when (1) fails in some way.

AFg6K7h4fhy2:
Description of Option 4, as requested. 

Option 4 receives two parameters. The first is an InferenceData object that is expected to have some groups (typically `observed_data` and `posterior_predictive`) associated with it. These groups might have one to several variables associated with them. For example, `posterior_predictive` might the variables `rt` and `obs`. The second parameter is a dictionary that has as its keys `str` corresponding to the groups in the `idata` and as its values dictionaries that map a dimension of a group's variable to a start date and a time interval. The start date is an ISO8601 `str` and the time interval is a `timedelta` expression from `datetime`. An example time interval could be `timedelta(weeks=2)`, meaning the coordinates of the group's variable's dimension will be start from the start date and be incremented by 2 weeks for a duration equal to size of the original coordinates.

Here is an example of how Option 4 is used. 

```python
option_4_idata_w_dates = option_4_add_dates_as_coords_to_idata(
    idata_wo_dates=idata_wo_dates_new,
    group_variable_date_mapping={
        "observed_data": {
            "obs": (
                "2022-08-08",
                timedelta(days=14),
                "obs_dim_0",
            ),
        },
        "posterior_predictive": {
            "obs": (
                "2022-08-08",
                timedelta(days=7),
                "obs_dim_0",  # change to f"{var_name}_dim_0"?
            ),
            "obs2": (
                "2022-08-08",
                timedelta(weeks=2),
                "obs2_dim_0",
            ),
            "rt": (
                "2022-08-01",
                timedelta(weeks=1),
                "rt_dim_0",
            ),
        },
    },
)
```

Here is the nature of the `idata` object that Option 4 receives:

```python
# in
print(idata_wo_dates_new["posterior_predictive"]["obs2"].coords)
print(idata_wo_dates_new["posterior_predictive"]["obs"].coords)
print(idata_wo_dates_new["posterior_predictive"]["rt"].coords)
# out (see next block)
```

```
Coordinates:
  * chain       (chain) int64 8B 0
  * draw        (draw) int64 8kB 0 1 2 3 4 5 6 7 ... 993 994 995 996 997 998 999
  * obs2_dim_0  (obs2_dim_0) int64 80B 0 1 2 3 4 5 6 7 8 9
Coordinates:
  * chain      (chain) int64 8B 0
  * draw       (draw) int64 8kB 0 1 2 3 4 5 6 7 ... 993 994 995 996 997 998 999
  * obs_dim_0  (obs_dim_0) int64 224B 0 1 2 3 4 5 6 7 ... 21 22 23 24 25 26 27
Coordinates:
  * chain     (chain) int64 8B 0
  * draw      (draw) int64 8kB 0 1 2 3 4 5 6 7 ... 993 994 995 996 997 998 999
  * rt_dim_0  (rt_dim_0) int64 96B 0 1 2 3 4 5 6 7 8 9 10 11
```

Once option 4 is used, the following is returned


```python
# in 
print(option_4_idata_w_dates["posterior_predictive"]["obs2"].coords)
print(option_4_idata_w_dates["posterior_predictive"]["obs"].coords)
print(option_4_idata_w_dates["posterior_predictive"]["rt"].coords)
# out (see below)
```

```
Coordinates:
  * chain       (chain) int64 8B 0
  * draw        (draw) int64 8kB 0 1 2 3 4 5 6 7 ... 993 994 995 996 997 998 999
  * obs2_dim_0  (obs2_dim_0) datetime64[ns] 80B 2022-08-08 ... 2022-12-12
Coordinates:
  * chain      (chain) int64 8B 0
  * draw       (draw) int64 8kB 0 1 2 3 4 5 6 7 ... 993 994 995 996 997 998 999
  * obs_dim_0  (obs_dim_0) datetime64[ns] 224B 2022-08-08 ... 2023-02-13
Coordinates:
  * chain     (chain) int64 8B 0
  * draw      (draw) int64 8kB 0 1 2 3 4 5 6 7 ... 993 994 995 996 997 998 999
  * rt_dim_0  (rt_dim_0) datetime64[ns] 96B 2022-08-01 2022-08-08 ... 2022-10-17
```

damonbayer:
thanks @AFg6K7h4fhy2! What do you think about making it easier to specify multiple variables that should get the same coordinates? For example there is a variable `X ` that exists in both the `prior` and `posterior` groups. Can we make it easy to assign the same coordinates to these variables?

Similarly, say there is `Rt` and `Rt_adj`, both in the `posterior` group. Can we make it easy to assign them the same coordinates?

AFg6K7h4fhy2:
Re: <https://github.com/CDCgov/forecasttools-py/pull/35#issuecomment-2448474285>

Yes, this should be possible.

Pre-implementation Plan: My immediate idea is to take this pattern 

```python
"<var_name>": (
    "<IS08601 str start date>",
    timedelta(<days or weeks>),
    "<dim_name>",
)
```

and change it to this pattern (or one similar to it):

```python
"<dim_name>": (
    "<IS08601 str start date>",
    timedelta(<days or weeks>),
    [<list of var_names associated with dim_name>],
)
```

damonbayer:
When are there multiple `var_name`s associated with a single `dim_name`?

AFg6K7h4fhy2:
Re: <https://github.com/CDCgov/forecasttools-py/pull/35#issuecomment-2448488404>

The `posterior_predictive` and `observed_data` groups of the `idata` objects packaged in `forecasttools` both have `obs_dim_0` as the dimension to their variables `obs` (`obs` is a variable of both groups).

dylanhmorris:
I think it might be productive both conceptually and from an implementation perspective to take a step back.

Currently, all the code I see is a single function that iterates over multiple `group`s / `var_name`s / etc. I think it will be more modular, testable, and readable for any such function to call, internally, a helper function that adds time coordinates to a single `var`. Good APIs for something that does lots of stuff at a pass will be easier to reason about in the context of this helper function and its API.

AFg6K7h4fhy2:
Re: <https://github.com/CDCgov/forecasttools-py/pull/35#issuecomment-2448507281>

Acknowledged. In addressing <https://github.com/CDCgov/forecasttools-py/pull/35#issuecomment-2448474285> I will take modularizing variable-level edits into account.

AFg6K7h4fhy2:
This comment (and subsequent ones) covers possible ways of addressing aforementioned comments: <https://github.com/CDCgov/forecasttools-py/pull/35#issuecomment-2448507281> and <https://github.com/CDCgov/forecasttools-py/pull/35#issuecomment-2448474285>.

The following structure is a possibility. This structure would permit multiple variables within a group to have the same coordinates. This function does not permit changing `obs`, which occurs in both groups specified, simultaneously. In the code examples provided in this PR (in [this file](https://github.com/CDCgov/forecasttools-py/pull/35/files#diff-d565a74e65206eabe5984b44c4f092edd21942e787ff8233b1b1f1641c740763)), `obs` is different across the `observed_data` and `posterior_predictive` groups: the latter group's `obs` has `forecast_days = 28` more values.


```python
idata_w_dates = add_dates_as_coords_to_idata(
    idata_wo_dates=idata_wo_dates,
    group_variable_date_mapping={
        "observed_data": {
            "obs": ("2022-08-08", timedelta(days=14), "obs_dim_0"),
        },
        "posterior_predictive": {
            ("obs", "obs2"): ("2022-08-08", timedelta(weeks=2), "obs_dim_0"),
            ("rt", "rt_adj"): ("2022-08-01", timedelta(weeks=1), "rt_dim_0"),
        },
    },
)
```

As for a helper function that permits changes to groups at the variable-level, the following is a possibility for its header:


```python
def add_time_coords_to_variable(
    group: xr.DataSet,
    variable_name: str,
    start_date_iso: str,
    time_step: datetime.timedelta,
    dim_name: str,
) -> xr.Dataset:
    pass
```

AFg6K7h4fhy2:
A call to a function that applies edits to variables across groups could look like:

```python
updated_idata = add_dates_across_groups(
    idata=idata_wo_dates,
    groups=["posterior_predictive", "observed_data"],
    variable_date_mapping=variable_date_mapping = {
        "obs": ("2022-08-08", timedelta(days=14), "obs_dim_0"),  
        "obs2": ("2022-08-08", timedelta(weeks=1), "obs2_dim_0"), 
        "rt": ("2022-08-01", timedelta(weeks=1), "rt_dim_0"),   
    }
)
```

An issue, however, occurs when the same variable name has different coordinates or dimension names across different groups. I believe need to assume some things about the naming of the dimensions. The following might be a remedy that is not as aesthetically pleasing but has the information to get the Goal completed:

```python
updated_idata = add_dates_across_groups(
    idata=idata_wo_dates,
    groups=["posterior_predictive", "observed_data"],
    variable_date_mapping= {
        "obs": {
            "observed_data": ("2022-08-08", timedelta(days=14), "obs_dim_0"),  
            "posterior_predictive": ("2022-08-08", timedelta(weeks=2), "obs_post_dim"),  
        },
        "obs2": {
            "posterior_predictive": ("2022-08-08", timedelta(days=7), "obs2_post_dim"),
        },
        "rt": {
            "all": ("2022-08-01", timedelta(weeks=1), "rt_dim_0"), # change across all groups
        },
    }
)
```

AFg6K7h4fhy2:
Which option is most robust, coordinates as 

* ISO8601 `str`
* `numpy.datetime`
* `datetime.datetime`?

damonbayer:
I suggest 

```python
def add_time_coords_to_variable(
    dat: xr.Dataset,  
    group: str,
    variable_name: str,
    dim_name: str,
    start_date_iso: str,
    time_step: datetime.timedelta,
) -> xr.Dataset:
    pass
```
Ideally this modifies the idata in place.

Alternatively, there could also be some class that contains `(group, variable_name, dim_name)`, and you could pass a list of those to `add_time_coords_to_variable`. That function would just loop through each element of the list and apply the correct coordinates based on `start_date_iso` and `time_step`.

AFg6K7h4fhy2:
Would the following function 

```python
def add_time_coords_to_variable(
    dat: xr.Dataset,  
    idata_locs = [(group, variable_name, dim_name), (group, variable_name, dim_name), …]
    start_date_iso: str,
    time_step: datetime.timedelta,
) -> xr.Dataset:
    pass
```

exist in addition to the first function (which modifies `idata` in place and which would be named something other than `add_time_coords_to_variable`)? If not, then the following function fails cases where different triples `(group, variable_name, dim_name)` require different start dates and time steps. This pattern would work if all variables and coordinates date intervals were unique, i.e. each quintuple `(group, variable_name, dim_name, start_date_iso, time_step)` is fully unique.

damonbayer:
> then the following function fails cases where different triples `(group, variable_name, dim_name)` require different start dates and time steps.

Assigning a single start date and time step to many variables is accomplished with one function call. If you need to assign a different start date and time step to some other variables, you can just use the function again.

AFg6K7h4fhy2:
The original suggestion 

```python
def add_time_coords_to_variable(
    dat: xr.Dataset,  
    group: str,
    variable_name: str,
    dim_name: str,
    start_date_iso: str,
    time_step: datetime.timedelta,
) -> xr.Dataset:
    pass
```

seems more reliable to me than passing the extracted group as an `xr.Dataset` and not having to pass the `az.InferenceData` object itself, assuming `dat: xr.Dataset` is supposed to be `dat: az.InferenceData`. Worth noting that `observed_data_group` in the following is an `xr.Dataset`, so `dat` might not be needed and group could be typed `xr.Dataset`. `dat`, as it stands in the above function, might not have a `group` attribute. 

```python
observed_data_group = getattr(idata_wo_dates, "observed_data", None)
print(type(idata_group))
```

AFg6K7h4fhy2:
> > then the following function fails cases where different triples `(group, variable_name, dim_name)` require different start dates and time steps.
> 
> Assigning a single start date and time step to many variables is accomplished with one function call. If you need to assign a different start date and time step to some other variables, you can just use the function again.

Ah, I see. With this pattern of usage, my preference rests with this function. Let me know if you have any other thoughts on this.

AFg6K7h4fhy2:
A discussion between {DHM, DB, TM} has designated some previous comments by TM as distracting. These comments are now struck-through. The outcome of the meeting was a plan for TM to develop an atomic function that modifies the coordinates of single variable before developing a function (which uses this atomic function) that processes multiple variables within the same group.

AFg6K7h4fhy2:
Inspection by @dylanhmorris and @damonbayer of `add_time_coords_to_idata_variable` and `add_time_coords_to_idata_variables` is pending (find [here](https://github.com/CDCgov/forecasttools-py/pull/35/files#diff-7d5d491c8860cb873e5b5cad35ce2ff248626a8fe73db562342d17b5ef06454c))

To facilitate this inspection of this newer code, I provide procedures that would be followed by the user in two instances.

Example: One has variables `obs` and `rt` in the `posterior_predictive` group of one's `idata` without dates. Both `obs` and `rt` have the same start date but `obs` follows daily dating and `rt` follows weekly dating. 

Response:

```python
# call 1 
idata_w_dates = add_time_coords_to_idata_variable(
    idata=idata_wo_dates,
    group="posterior_predictive",
    variable="obs",
    dimensions=["obs_dim_0"],
    start_date_iso="2022-08-08",
    time_step=timedelta(days=1),
)

# call 2
idata_w_dates = add_time_coords_to_idata_variable(
    idata=idata_wo_dates,
    group="posterior_predictive",
    variable="rt",
    dimensions=["rt_dim_0"],
    start_date_iso="2022-08-08",
    time_step=timedelta(weeks=1),
)
```

Example: One has variables named `obs` in the `observed_data` group and in the `posterior_predictive` group in one's `idata` without dates. Both `obs`s have the same start date and follow daily dating. However, the coordinate sizes differ across groups, with `obs` in `observed_data` being 488 units and `obs` in `posterior_predictive` being 516 units.

Response:

```python
idata_w_dates = add_time_coords_to_idata_variables(
    idata=idata_wo_dates,
    group_var_dim_tuples=[
        ("posterior_predictive", "obs", "obs_dim_0"),
        ("observed_data", "obs", "obs_dim_0"),
    ],
    start_date_iso="2022-08-08",
    time_step=timedelta(weeks=1),
)
```

Example: One has variables `obs` and `rt` in the `posterior_predictive` group of one's `idata` without dates. Both `obs` and `rt` follow daily dating, but `obs` starts one week earlier than `rt`. 

Response:


```python
# call 1 
idata_w_dates = add_time_coords_to_idata_variable(
    idata=idata_wo_dates,
    group="posterior_predictive",
    variable="obs",
    dimensions=["obs_dim_0"],
    start_date_iso="2022-08-08",
    time_step=timedelta(days=1),
)

# call 2
idata_w_dates = add_time_coords_to_idata_variable(
    idata=idata_wo_dates,
    group="posterior_predictive",
    variable="rt",
    dimensions=["rt_dim_0"],
    start_date_iso="2022-08-01",
    time_step=timedelta(days=1),
)
```

AFg6K7h4fhy2:
Should the functions `add_time_coords_to_idata_variable` and `add_time_coords_to_idata_variables` from the [last commit](https://github.com/CDCgov/forecasttools-py/pull/35/commits/7709489d588c62d4fae2c6595d5107fbfef37809) be retained fully or partially, then these functions remain to be added from the notebook <https://github.com/CDCgov/forecasttools-py/pull/35/files#diff-7d5d491c8860cb873e5b5cad35ce2ff248626a8fe73db562342d17b5ef06454c> to `forecasttools`. 

In expectation (i.e. bottlenecked by the approval of the aforementioned functions), the function `add_dates_as_coords_to_idata` (see [here](https://github.com/CDCgov/forecasttools-py/blob/558c64009572b2f9e8100dc436480d2e1b358901/forecasttools/idata_w_dates_to_df.py#L17C5-L17C33)) will be replaced and the function `idata_forecast_w_dates_to_df` (which is, similar to `add_dates_as_coords_to_idata`, contained in `./forecasttools/idata_w_dates_to_df.py`) will have to be modified to support the new utilities.

notebooks/test_idata_general_time_representation.py
```diff
@@ -0,0 +1,237 @@
...
+        the group and variable to modify.
+    group : str
+        The name of the group within the InferenceData
+        object (e.g., "posterior_predictive").
+    variable : str
+        The name of the variable within the
+        specified group to assign time coordinates to.
+    dimensions : list[str]
+        A list of dimension names to which time
+        coordinates should be assigned.
```
dylanhmorris, line 57: Should this default to the zeroth dimension if not specified to save the user typing?

notebooks/test_idata_general_time_representation.py
```diff
@@ -0,0 +1,237 @@
...
+    start_date_iso : str
+        The start date for the time coordinates
+        in ISO format (e.g., "2022-08-08").
```
dylanhmorris, line 60: accept either a string or a `datetime.date`?

dylanhmorris: Or even just a `datetime.date` if you're going to require the next argument to be a `timedelta` (as opposed to, e.g., a parseable string)

notebooks/test_idata_general_time_representation.py
dylanhmorris: Also, is there any instance in which you'd want to assign the same timeseries as the coordinate for more than one dimension?

notebooks/test_idata_general_time_representation.py
```diff
@@ -0,0 +1,237 @@
...
+            f"Variable '{variable}' not found in group '{group}'."
+        )
+    # retrieve the variable's data array
+    variable_data = idata_group[variable]
+    # check and apply time coordinates only to specified dimensions that exist in the variable
+    applicable_dims = [dim for dim in dimensions if dim in variable_data.dims]
+    if not applicable_dims:
+        raise ValueError(
+            f"No specified dimensions found in variable '{variable}' within group '{group}'."
+        )
```
dylanhmorris, line 95: As above, I'm not sure this should ever assign more than one dimension at a time, at least as written/implemented.

notebooks/test_idata_general_time_representation.py
AFg6K7h4fhy2: Does the option (`start_date_iso` as `str` or `datetime.date` AND `time_steps` as `str` or `datetime.timedelta`) seem better to you than (`start_date_iso` as `datetime.date` and `time_steps` `datetime.timedelta`)?

notebooks/test_idata_general_time_representation.py
AFg6K7h4fhy2: `chain` and `draws` are dimensions, so maybe the default behavior could be first non-chain, non-draw dimension if dimension(s) are not specified?

AFg6K7h4fhy2: Re: <https://github.com/CDCgov/forecasttools-py/pull/35#discussion_r1828562332>

Yes, there are such instances. There are sometimes dimensions across variables and (perhaps also) within variables that we'd want to have the same time coordinates. I gave an example of this earlier (with the minor difference that the two had different end dates; the "sameness" comes from their having the same start date and date interval) (emphasis has been added in this message):

> Example 2: One has variables named `obs` in the `observed_data` group and in the `posterior_predictive` group in one's `idata` without dates. **Both `obs`s have the same start date and follow daily dating.** However, the coordinate sizes differ across groups, with `obs` in `observed_data` being 488 units and `obs` in `posterior_predictive` being 516 units.
> 
> Response:

```python
idata_w_dates = add_time_coords_to_idata_variables(
    idata=idata_wo_dates,
    group_var_dim_tuples=[
        ("posterior_predictive", "obs", "obs_dim_0"),
        ("observed_data", "obs", "obs_dim_0"),
    ],
    start_date_iso="2022-08-08",
    time_step=timedelta(days=1),
)
```

dylanhmorris: > Also, is there any instance in which you'd want to assign the same timeseries as the coordinate for more than one dimension?

Clarification: for more than one dimension _of the same variable_

notebooks/test_idata_general_time_representation.py
dylanhmorris: I think my preference is `start_date` as `str` or `datetime.date` and time_step as `datetime.timedelta` is my preference

notebooks/test_idata_general_time_representation.py
AFg6K7h4fhy2: Sounds good. I will go with this preference.

notebooks/test_idata_general_time_representation.py
AFg6K7h4fhy2: Sounds good; I will go with this preference.

notebooks/test_idata_general_time_representation.py
dylanhmorris: example?

forecasttools/__init__.py
AFg6K7h4fhy2, line 1: This PR #33 does away with relative imports.

AFg6K7h4fhy2:
Would appreciate another inspection here as well @dylanhmorris .

forecasttools/__init__.py
dylanhmorris, line 1: Worth discussing whether all functions should be exposed via the top-level module, but out of scope for now.

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,245 @@
...
+            )
+    elif isinstance(start_date_iso, datetime):
+        start_date_as_dt = start_date_iso
+    else:
+        raise TypeError(
+            f"Parameter 'start_date_iso' must be of type 'str' or 'datetime'; got {type(start_date_iso)}"
+        )
+    if not isinstance(time_step, timedelta):
+        raise TypeError(
+            f"Parameter 'time_step' must be of type 'datetime.timedelta'; got {type(time_step)}"
```
dylanhmorris, line 101: This should be DRY-ified and/or omitted and handled with duck typing.

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,245 @@
...
+    # to the specified dimensions that exist
+    # in the variable
+    if dimension not in variable_data.dims:
+        raise ValueError(
+            f"Dimension '{dimension}' not found in variable dimensions: '{variable_data.dims}'."
+        )
+    # determine the interval size for the
+    # selected dimension
+    interval_size = variable_data.sizes[dimension]
+    # generate date range using the specified time_step and interval size
```
dylanhmorris, line 127: Comment only when the code is opaque. See the [predict programming guidelines](https://potential-adventure-637em36.pages.github.io/docs/style_guide.html#documentation)

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,245 @@
...
+    interval_dates = (
+        pl.date_range(
+            start=start_date_as_dt,
+            end=start_date_as_dt + (interval_size - 1) * time_step,
+            interval=f"{time_step.days}d" if time_step.days else "1d",
```
dylanhmorris, line 132: This will silently round all timedeltas that do not represent an integer number of days down to the nearest day. I am not convinced that is desirable.

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,245 @@
...
+    if not isinstance(time_step, timedelta):
+        raise TypeError(
+            f"Expected 'time_step' to be a 'timedelta'; got {type(time_step)}."
+        )
+    # check that all groups, variables, and dimensions are strings
+    if not all(isinstance(g, str) for g in groups):
+        raise TypeError("All items in 'groups' must be strings.")
+    if not all(isinstance(v, str) for v in variables):
+        raise TypeError("All items in 'variables' must be strings.")
+    if not all(isinstance(d, str) for d in dimensions):
```
dylanhmorris, line 214: Again, DRYify or remove

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,245 @@
...
+        raise TypeError("All items in 'dimensions' must be strings.")
+    # iterate over (group, variable, dimension) triples
+    for group, variable, dimension in zip(groups, variables, dimensions):
+        try:
```
dylanhmorris, line 218: I am not convinced this should be a try catch, but I am persuadable. Is the idea that it's useful to know where in the loop things errored?

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,245 @@
...
+                start_date_iso=start_date_iso,
+                time_step=time_step,
             )
-            setattr(
-                idata_w_dates,
-                group_name,
-                idata_group_with_dates,
+        except ValueError as e:
+            print(
+                f"Error for (group={group}, variable={variable}, dimension={dimension}): {e}"
```
dylanhmorris, line 245: Better practice to include this directly in the error message via the ["raise from" syntax](https://stackoverflow.com/questions/24752395/python-raise-from-usage)

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,245 @@
...
-                .astype("datetime64[ns]")
-            )
-            idata_group_with_dates = idata_group.assign_coords(
-                {dim_name: interval_dates}
+            # check if the dimension exists
+            variable_data = group_data[variable]
+            if dimension not in variable_data.dims:
+                raise ValueError(
+                    f"Dimension '{dimension}' not found for variable '{variable}' in group '{group}'."
+                )
```
dylanhmorris, line 233: Doesn't `add_time_coords_to_idata_dimension` perform this check?

forecasttools/__init__.py
AFg6K7h4fhy2: Discussed in issue #39 .

forecasttools/utils.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def validate_and_get_idata_group(idata: az.InferenceData, group: str):
+    """Retrieves the group from the
+    InferenceData object and validates its
+    existence."""
+    idata_group = getattr(idata, group, None)
+    if idata_group is None:
+        raise ValueError(f"Group '{group}' not found in idata object.")
+    return idata_group
```
dylanhmorris, line 68: I think I prefer 
```suggestion
    if not hasattr(idata, group):
        raise ValueError(f"Group '{group}' not found in idata object.")
    return getattr(idata, group)
```
this is also more consistent with the pattern in `validate_and_get_idata_group_var()` below

forecasttools/utils.py
```diff
@@ -0,0 +1,183 @@
...
+def validate_idata_group_var_dim(variable_data: xr.DataArray, dimension: str):
+    """Validates if the given dimension
+    exists in the variable."""
+    if dimension not in variable_data.dims:
+        raise ValueError(
+            f"Dimension '{dimension}' not found in variable dimensions: '{variable_data.dims}'."
+        )
+
+
+def validate_group_var_dim_instances(
```
dylanhmorris, line 94: Name this something more generic?

forecasttools/utils.py
```diff
@@ -0,0 +1,183 @@
...
+    instances of expected_type. If not, raise
+     a TypeError.
+    """
+    if not all(isinstance(item, expected_type) for item in iterable):
+        raise TypeError(
+            f"All items in '{param_name}' must be of type '{expected_type.__name__}'."
+        )
+
+
+def generate_time_range_for_dim(
```
dylanhmorris, line 108: If it's specific to an idata dimension, which it currently is, this belongs in `idata_w_tdates_to_df`, not `utils`

forecasttools/utils.py
```diff
@@ -0,0 +1,183 @@
...
+def ensure_listlike(x):
+    """
+    Ensure that an object either behaves like a
+    :class:`MutableSequence` and if not return a
+    one-item :class:`list` containing the object.
+    Useful for handling list-of-strings inputs
+    alongside single strings.
+    Based on this _`StackOverflow approach
+    <https://stackoverflow.com/a/66485952>`.
+    Parameters
```
dylanhmorris, line 172:
```suggestion

    Parameters
```

tests/test_add_time_coords_to_idata.py
```diff
@@ -0,0 +1,441 @@
...
+        "obs",
+        "obs_dim_0",
+        start_date_iso,
+        time_step,
+    )
+    assert "obs_dim_0" in idata.posterior_predictive["obs"].coords
+    assert (
+        len(idata.posterior_predictive["obs"].coords["obs_dim_0"])
+        == idata.posterior_predictive["obs"].sizes["obs_dim_0"]
+    )
```
dylanhmorris, line 113: These seem now to be covered by the test above, provided you add another parametrization with a valid datetime `start_date`

tests/test_add_time_coords_to_idata.py
```diff
@@ -0,0 +1,441 @@
...
+        # validate that the function executed
+        # correctly and the dimension has been modified
+        assert dimension in idata_out.posterior_predictive[variable].coords
+        assert (
+            len(idata_out.posterior_predictive[variable].coords[dimension])
+            == idata_out.posterior_predictive[variable].sizes[dimension]
+        )
+        # old_dim = idata[group][variable][dimension]
+        # new_dim = idata_out[group][variable][dimension]
+        # assert not np.array_equal(old_dim, new_dim)
```
dylanhmorris, line 272: Why is this commented out?

tests/test_add_time_coords_to_idata.py
dylanhmorris, line 1: Important to have at least one test in which we check that the output is as expected (i.e. dates added with the utility correspond to a set of expected dates hard coded or calculated manually).

dylanhmorris (changes requested):
Thanks, @AFg6K7h4fhy2. Largely ready, but a few remaining things.

forecasttools/idata_w_dates_to_df.py
```diff
@@ -181,7 +181,8 @@ def add_time_coords_to_idata_dimensions(
         except ValueError as e:
             raise ValueError(
                 f"Error for (group={group}, variable={variable}, dimension={dimension}): {e}"
```
dylanhmorris, line 240: If you do this, you don't need the `{e}`

AFg6K7h4fhy2:
In working on the demonstration, which will be removed shortly from this branch, 2 bugs were found. One bug does not directly concern the changes made in this PR; as such, it has been made into an issue (#42 ). The other bug, however, does concern this PR. The following, from the demonstration, does not produce `datetime[D]` and instead produces `datetime[ns]`, which it should not, since `timedelta(days=1)` evenly divides the number of seconds in a day. 

```markdown
```{python}
idata_wo_dates = forecasttools.nhsn_flu_forecast_wo_dates

idata_wo_dates

idata_wo_dates["posterior_predictive"]["obs"]["obs_dim_0"]
```

__Intermediary Functions For Adding Dates Or Time__

```{python}
start_date_iso = "2022-08-01"

start_time_as_dt = forecasttools.validate_and_get_start_time(start_date_iso)

variable_data = idata_wo_dates["posterior_predictive"]["obs"]

as_dates_idata = forecasttools.generate_time_range_for_dim(
    start_time_as_dt=start_time_as_dt,
    variable_data=variable_data,
    dimension="obs_dim_0",
    time_step=timedelta(days=1),
)
print(as_dates_idata[:5], type(as_dates_idata[0]))

as_time_idata = forecasttools.generate_time_range_for_dim(
    start_time_as_dt=start_time_as_dt,
    variable_data=variable_data,
    dimension="obs_dim_0",
    time_step=timedelta(days=1.5),
)
print(as_time_idata[:5], type(as_time_idata[0]))
```

__Adding Dates To Single Versus Multiple Groups__

```{python}
idata_w_dates_single = forecasttools.add_time_coords_to_idata_dimension(
    idata=idata_wo_dates,
    group="posterior_predictive",
    variable="obs",
    dimension="obs_dim_0",
    start_date_iso=start_date_iso,
    time_step=timedelta(weeks=1), # notice weeks
) 
idata_w_dates_single.posterior_predictive
```

AFg6K7h4fhy2:
Re: <https://github.com/CDCgov/forecasttools-py/pull/35#issuecomment-2486043198>

The action pathway here involves refactoring `generate_time_range_for_dim` to accept either `datetime.date` or `datetime.datetime`. If the `time_step` is an integer number of days or weeks, then the `pl.date_range` is used and other `pl.datetime_range` is used.

AFg6K7h4fhy2:
A quick note from personal conversation with DHM.

`forecasttools-py` will no longer accept `str` input for `start_date_iso`. If one wants to have dates, `datetime.date(2024, 11, 20)` is not much more effort than `2024-11-20` (and would usually just be written as `date(2024, 11, 20)`).

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,226 @@
...
+    and/or the time step is a time.
+    """
+
+    # get the size of the dimension
+    interval_size = variable_data.sizes[dimension]
+
+    # check if time_step is in full days or
+    # weeks
+    is_full_days_or_weeks = (
+        time_step.total_seconds() % timedelta(days=1).total_seconds() == 0.0
```
dylanhmorris, line 42: I think I prefer a helper function, and one that avoids float comparison with 0. Something like

```python
def is_integer_days(td: datetime.timedelta):
    if td.resolution > datetime.timedelta(days=1):
        raise ValueError("Cannot check whether a timedelta "
                         "represents an integer number of days "
                         "for a timedelta with resolution greater than 1 day. "
                         f"Got {td.resolution}")
    return abs(td -  td.days) < t.resolution
```

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,226 @@
...
+            eager=True,
+        ).to_list()
+    else:
+        # use pl.datetime_range for datetime
+        return pl.datetime_range(
+            start=start_time_as_dt,
+            end=start_time_as_dt + (interval_size - 1) * time_step,
+            interval=time_step,
+            closed="both",
+            eager=True,  # return a Polars series
```
dylanhmorris, line 62:
```suggestion
            eager=True,
```

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,226 @@
...
+        )
+    idata_group = forecasttools.validate_and_get_idata_group(
+        idata=idata, group=group
+    )
+    variable_data = forecasttools.validate_and_get_idata_group_var(
+        idata_group=idata_group, group=group, variable=variable
+    )
+    forecasttools.validate_idata_group_var_dim(
+        variable_data=variable_data, dimension=dimension
+    )
```
dylanhmorris, line 136: Beyond the scope of this PR, but once this lands there should be a cleanup PR to remove these validation helpers from the top-level namespace. Even if they're going to be user-exposed, that's too prominent I think.

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,226 @@
...
+        The modified InferenceData object with
+        updated time coordinates for the
+        specified groups, variables, and
+        dimensions.
+    """
+    inputs = [
+        (idata, az.InferenceData, "idata"),
+        (groups, (str, list), "groups"),
+        (variables, (str, list), "variables"),
+        (dimensions, (str, list), "dimensions"),
```
dylanhmorris, line 196: Again, I think this can be duck-typed, especially as you're going to call `ensure_listlike` on them

dylanhmorris: Why error if the user provides a `numpy.array` of strings, for instance?

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,226 @@
...
+    groups = forecasttools.ensure_listlike(groups)
+    variables = forecasttools.ensure_listlike(variables)
+    dimensions = forecasttools.ensure_listlike(dimensions)
+    # check groups, variables, and dimensions
+    # all contain str vars
+    forecasttools.validate_iter_has_expected_types(groups, str, "groups")
+    forecasttools.validate_iter_has_expected_types(variables, str, "variables")
+    forecasttools.validate_iter_has_expected_types(
+        dimensions, str, "dimensions"
+    )
```
dylanhmorris, line 212: Again, this feels like unnecessarily strong typing for a duck-y language, but I am less opposed here than I am above.

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,226 @@
...
+    # iterate over (group, variable, dimension) triples
+    for group, variable, dimension in zip(groups, variables, dimensions):
+        try:
+            idata = add_time_coords_to_idata_dimension(
+                idata=idata,
+                group=group,
+                variable=variable,
+                dimension=dimension,  # validated in called func
+                start_date_iso=start_date_iso,  # validated in called func
+                time_step=time_step,  # validated in called func
```
dylanhmorris, line 222:
```suggestion
                dimension=dimension,
                start_date_iso=start_date_iso,
                time_step=time_step,
```

tests/test_add_time_coords_to_idata.py
```diff
@@ -0,0 +1,477 @@
...
+        dimension=dimension,
+        start_date_iso=start_date_iso,
+        time_step=time_step,
+    )
+
+    # extract the updated time coordinates
+    updated_coords = updated_idata.observed_data.coords[dimension].values
+
+    # assert that the updated coordinates are different from expected dates
+    assert any(a == b for a, b in zip(updated_coords, expected_dates))
```
dylanhmorris, line 308: 1. fix comment
2. Should be an all, not an any, no?
3. Why not use `np.testing.assert_array_equal`?
```suggestion
    # assert that the updated coordinates are equal to the expected dates
    assert all(a == b for a, b in zip(updated_coords, expected_dates))
```

dylanhmorris: Also, we should have a similar result == expected test where the result is times rather than dates (i.e. both when the timedelta is an integer number of days but the start date is a full datetime and when the start date is a date but the timedelta is not an integer date)

forecasttools/idata_w_dates_to_df.py
AFg6K7h4fhy2: Your note here and in [this comment](https://github.com/CDCgov/forecasttools-py/pull/35#discussion_r1852584560) has been added to this issue #41 (see [this comment](https://github.com/CDCgov/forecasttools-py/issues/41#issuecomment-2491891359)).

tests/test_add_time_coords_to_idata.py
AFg6K7h4fhy2: You're correct. If you had looked earlier today, I any `any` and `a != b`, then I reacted, without thinking much, believing I was incorrect and changed it from `a != b` to `a == b`. Good catch.

AFg6K7h4fhy2: I had `np.testing.assert_array_equal` when `pl.date_range` and `pl.datetime_range` were cast to `numpy`. But now, to get past issues with `numpy.datetime64[D]` and `numpy.datetime64[ns]`, which are introduced in using `numpy`, I have `to_list()` rather than `to_numpy()`, so I am no longer using `np.testing.assert_array_equal()` or `assert not np.array_equal()`.

dylanhmorris: Indeed

```python
from numpy.testing import assert_array_equal
assert_array_equal([datetime.date(2024,1,1), datetime.date(2022,3,5)], [datetime.date(2024,1,2), datetime.date(2022,3,5)])  # raises AssertionError
```

```python
assert_array_equal([datetime.date(2024,1,1), datetime.date(2022,3,5)], [datetime.date(2024,1,1), datetime.date(2022,3,5)]) # no error
```

forecasttools/idata_w_dates_to_df.py
AFg6K7h4fhy2: I am in the process of adding something like this.

tests/test_add_time_coords_to_idata.py
dylanhmorris: This is an example of duck typing working as it should. `assert_array_equal` can be used on lists (and jax arrays, and other things), not just on numpy arrays. (None of which is to say that there aren't very real drawbacks to duck typing / advantages of stricter typing)

AFg6K7h4fhy2:
In communications between DHM and the author, a decision has been made to cease using `pl.date_range` and `pl.datetime_range` and instead use `np.arange` with utilities for `datetime64[ns]` and `datetime64[D]` objects.

AFg6K7h4fhy2:
The author's remark:

> Even if one uses the approach involving np.arange w/ timedelta64 "D", assign_coords converts to [ns] precision, keeping the "times" even if one is operating in "dates". Bite the bullet here an just always accept "times" for coords? I want to canvas assign_coords once more.

```python
from datetime import date, datetime, timedelta

import arviz as az
import numpy as np
import polars as pl
import xarray as xr


coords = {"obs_dim_0": [0, 1, 2, 3, 5]}
data = xr.Dataset(
    {"obs": ("obs_dim_0", np.array([10, 20, 30, 40, 50]))}, coords=coords
)
idata = az.InferenceData(observed_data=data)

# test parameters
group = "observed_data"
variable = "obs"
dimension = "obs_dim_0"
start_date = date(2022, 8, 1)

expected_datetimes = np.array([
    datetime(2022, 8, 1),
    datetime(2022, 8, 2),
    datetime(2022, 8, 3),
    datetime(2022, 8, 4),
    datetime(2022, 8, 5),
]).astype("datetime64[D]")

idata_group = getattr(idata, group)
variable_data = idata_group[variable]
interval_size = variable_data.sizes[dimension]

out = np.arange(
    start_date, interval_size, np.timedelta64(1, "D")
)

print(out)

idata_group = idata_group.assign_coords({dimension: out})

setattr(idata, group, idata_group)

print(idata.observed_data.coords[dimension].values)
```

Out:

```
['2022-08-01' '2022-08-02' '2022-08-03' '2022-08-04' '2022-08-05']
['2022-08-01T00:00:00.000000000' '2022-08-02T00:00:00.000000000'
 '2022-08-03T00:00:00.000000000' '2022-08-04T00:00:00.000000000'
 '2022-08-05T00:00:00.000000000']
<ipython-input-2-6ff7ebed4ad4>:35: UserWarning: Converting non-nanosecond precision datetime values to nanosecond precision. This behavior can eventually be relaxed in xarray, as it is an artifact from pandas which is now beginning to support non-nanosecond precision values. This warning is caused by passing non-nanosecond np.datetime64 or np.timedelta64 values to the DataArray or Variable constructor; it can be silenced by converting the values to nanosecond precision ahead of time.
  idata_group = idata_group.assign_coords({dimension: out})
```

dylanhmorris:
Yes, if `xarray` currently only supports `[ns]` precision, we should do the same in this function.

AFg6K7h4fhy2:
Re: <https://github.com/CDCgov/forecasttools-py/pull/35#issuecomment-2492179877>

With this, the author will still operate with using two paths, one for `datetime64[D]` and another for `datetime64[ns]`. However, it will be noted in a docstring / type hint that the output is `datetime64[ns]`. Rather than remove what will operate as intended as some point in the future (i.e. when `xarray` behavior is relaxed w/ non `[ns]` precision), the author will retain the separation, which does not seem to harm usability.

dylanhmorris:
Sounds good, but to avoid the warning, you should coerce the `[D]` resolution arrays to `[ns]` resolution before calling `assign_coords()`

AFg6K7h4fhy2:
Re: <https://github.com/CDCgov/forecasttools-py/pull/35#issuecomment-2492218781>

Sounds good; I've been proceeding with properly placed `.astype("datetime64[ns]")`s.

forecasttools/idata_w_dates_to_df.py
```diff
@@ -17,15 +17,44 @@
...
-            "for a timedelta with resolution greater than 1 day. "
-            f"Got {td.resolution}"
+def is_timedelta_in_days_only(td: timedelta) -> bool:
+    """
+    Checks if a timedelta object can be
+    represented exclusively in days,
+    meaning it has no hours, minutes,
+    seconds, or microseconds.
+    """
+    return td.seconds == 0 and td.microseconds == 0
```
dylanhmorris, line 27: Nice idea. Much clearer.

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,250 @@
...
+    """
+    Converts a Python timedelta to:
+    numpy.timedelta64[D] if it is
+    representable in days only. Otherwise,
+    convert to numpy.timedelta64[ns].
+    """
+    if is_timedelta_in_days_only(td):
+        return np.timedelta64(td.days, "D")
+    else:
+        return np.timedelta64(int(td.total_seconds() * 1e9), "ns")
```
dylanhmorris, line 57: This feels unnecessarily complicated and potentially unsafe.  I think better simply to do
```suggestion
        return np.timedelta64(td).astype("timedelta64[ns]")
```

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,250 @@
...
+    time_step: timedelta,
+) -> list[date] | list[datetime]:
+    """
+    Generates a range of times based on the
+    start date, time step, and variable's
+    dimension size. A range of dates is
+    generated if the start date is a date and
+    the time step is in days. A range of times
+    is generated if the start date is a time
+    and/or the time step is a time.
```
dylanhmorris, line 73: Remove docstring text that is no longer accurate
```suggestion
    dimension size. 
```

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,250 @@
...
+    else:
+        return np.timedelta64(int(td.total_seconds() * 1e9), "ns")
+
+
+def generate_time_range_for_dim(
+    start_time_as_dt: datetime | date,
+    variable_data: xr.DataArray,
+    dimension: str,
+    time_step: timedelta,
+) -> list[date] | list[datetime]:
```
dylanhmorris, line 65:
```suggestion
) -> np.ndarray
```

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,250 @@
...
+    convert to numpy.timedelta64[ns].
+    """
+    if is_timedelta_in_days_only(td):
+        return np.timedelta64(td.days, "D")
+    else:
+        return np.timedelta64(int(td.total_seconds() * 1e9), "ns")
+
+
+def generate_time_range_for_dim(
+    start_time_as_dt: datetime | date,
```
dylanhmorris, line 61:
```suggestion
    start_time_as_dt: datetime | date | np.datetime64,
```

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,250 @@
...
+    variable_data: xr.DataArray,
+    dimension: str,
+    time_step: timedelta,
```
dylanhmorris, line 64:
```suggestion
    time_step: timedelta | np.timedelta64,
```

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,250 @@
...
+
+
+def convert_date_or_datetime_to_np(time_object: any) -> np.datetime64:
+    """
+    Converts a date or datetime object to
+    numpy.datetime64: date -> datetime64[D]
+    (day precision), datetime ->
+    datetime64[ns] (nanosecond precision).
+    """
+    if isinstance(time_object, date) and not isinstance(time_object, datetime):
```
dylanhmorris, line 37: Add option to return unchanged if argument _is_ an `np.datetime64`

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,250 @@
...
+        )
+
+
+def convert_timedelta_to_np(td: timedelta) -> np.timedelta64:
+    """
+    Converts a Python timedelta to:
+    numpy.timedelta64[D] if it is
+    representable in days only. Otherwise,
+    convert to numpy.timedelta64[ns].
+    """
```
dylanhmorris, line 53: Add option to return unchanged if argument _is_ an np.timedelta64

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,250 @@
...
+        step=time_step_as_np,
+    )
+
+
+def add_time_coords_to_idata_dimension(
+    idata: az.InferenceData,
+    group: str,
+    variable: str,
+    dimension: str,
+    start_date_iso: datetime | date,
```
dylanhmorris, line 95:
```suggestion
    start_date_iso: datetime | date | np.datetime,
```

forecasttools/idata_w_dates_to_df.py
```diff
@@ -7,80 +7,250 @@
...
+        time coordinates for the specified
+        group, variable, and dimension.
+    """
+    inputs = [
+        (idata, az.InferenceData, "idata"),
+        (group, str, "group"),
+        (variable, str, "variable"),
+        (dimension, str, "dimension"),
+        (time_step, timedelta, "time_step"),
+        (start_date_iso, (date, datetime), "start_date_iso"),
```
dylanhmorris, line 146: allow np.dattimes and timedeltas here

tests/test_add_time_coords_to_idata.py
```diff
@@ -0,0 +1,525 @@
...
+        variable=variable,
+        dimension=dimension,
+        start_date_iso=start_date_iso,
+        time_step=time_step,
+    )
+
+    # extract the updated time coordinates
+    updated_coords = updated_idata.observed_data.coords[dimension].values
+
+    print(updated_coords)
```
dylanhmorris, line 308:
```suggestion
```

tests/test_add_time_coords_to_idata.py
```diff
@@ -0,0 +1,525 @@
...
+    # expected output dates
+    expected_dates = np.array(
+        [
+            date(2024, 11, 19),
+            date(2024, 11, 21),
+            date(2024, 11, 23),
+            date(2024, 11, 25),
+            date(2024, 11, 27),
+        ]
+    ).astype("datetime64[D]")
```
dylanhmorris, line 293: should this be 
```suggestion
    ).astype("datetime64[ns]")
```

dylanhmorris (changes requested):
Looking very good, thanks @AFg6K7h4fhy2. A few final things.

tests/test_add_time_coords_to_idata.py
AFg6K7h4fhy2: Making a note that when `xarry` time precision is relaxed, this will change back to `[D]`.

dylanhmorris (approved):
Thanks for your hard work, @AFg6K7h4fhy2!

Issue #18, opened 2024-10-11
Function to cast `InferenceData` into `tidy_draws` format


damonbayer:
`tidy_draws` [details](https://mjskay.github.io/tidybayes/reference/tidy_draws.html).

You can approach this however you like, but I have previously done some relevant work:

I have written similar functionality in R, which expects as input the output of  `arviz.InferenceData.to_dataframe`.
It splits the `InferenceData` into several `tidy_draws` formatted tables, one for each group.

```r
arviz_split <- function(x) {
x %>%
    select(-distribution) %>%
    split(f = as.factor(x$distribution))
}

pyrenew_samples <-
read_csv(inference_data_path) %>%
rename_with(\(varname) str_remove_all(varname, "\\(|\\)|\\'|(, \\d+)")) |>
rename(
    .chain = chain,
    .iteration = draw
) |>
mutate(across(c(.chain, .iteration), \(x) as.integer(x + 1))) |>
mutate(
    .draw = tidybayes:::draw_from_chain_and_iteration_(.chain, .iteration),
    .after = .iteration
) |>
pivot_longer(-starts_with("."),
    names_sep = ", ",
    names_to = c("distribution", "name")
) |>
arviz_split() |>
map(\(x) pivot_wider(x, names_from = name) |> tidy_draws())
```

AFg6K7h4fhy2:
A comment is due on why this PR was modified from High Priority to Medium Priority during the week of [Oct. 28, Nov. 1]. Given DB's workflow (see partially [here](https://github.com/CDCgov/forecasttools-py/issues/18#issue-2581664532)), which loads in R the contents of an `az.InferenceData` object, in `pyrenew-hew` (see [here](https://github.com/CDCgov/pyrenew-hew)), careful treatment of time intervals in `az.InferenceData` was needed more (see [here](https://github.com/CDCgov/forecasttools-py/issues/34)) by DB, who with the rest of STF represents the primary user of `forecasttools-py`, for the time being.

AFg6K7h4fhy2:
In personal communication between the author and DHM, it was remarked that starting with reimplementing Damon's code (which the author has already been done) ought to be the first step for this task rather than the author's existing approach of converting _all_ groups, even possibly extraneous ones, from within `az.InferenceData` to be compatible with `tidydraws`.

damonbayer:
[...]

PR #36, opened 2024-10-28
Function To Cast InferenceData Into `tidy_draws` Format

O957:
For the scope of this PR, please refer to issue #18 .


notebooks/idata_to_tidy_draws.py
```diff
@@ -0,0 +1,456 @@
...
+    df = df.with_columns(
+        [
+            (pl.col("chain") + 1).alias(".chain"),  # add 1 for conversion to R
+            (pl.col("draw") + 1).alias(
+                ".iteration"
+            ),  # add 1 for conversion to R
+            (
+                (pl.col("chain") * df["draw"].n_unique()) + pl.col("draw") + 1
+            ).alias(
+                ".draw"
```
dylanhmorris, line 378: shouldn't this be `draw`?

notebooks/idata_to_tidy_draws.py
```diff
@@ -0,0 +1,456 @@
...
+        ]
+    ).drop(
+        ["chain", "draw"]
+    )  # drop original chain, draw
+    # NOTE: anything needed for cleaning for R col names?
+    # pivot to long format to have variables in a single column
+
+    # need to correct this further...
+    tidy_df = df.unpivot(
+        index=[".chain", ".iteration", ".draw"],
```
dylanhmorris, line 389: No dots I think
```suggestion
        index=["chain", "iteration", "draw"],
```

damonbayer: Why not? This is the convention in tidybayes https://mjskay.github.io/tidybayes/articles/tidy-brms.html#gathering-variable-indices-into-a-separate-column-in-a-tidy-format-data-frame

AFg6K7h4fhy2 (commented):
A quick self-review of this PR:

* `.csv` being ignored in `.gitignore` seems like a good practice.
* The `black` `pre-commit` hook will be fixed [^black].
* The R code in `./assets/external/` seems fine to me, but I can understand why others might not want these files in `forecasttools-py`. Middle ground might be a future document or README section that connects `forecasttools-py` implementations with those in `forecasttools` or `pyrenew`-based repositories.
* I believe citing software packages should be standard across CFA and CDCgov repositories. This explains the presence of `references.bib`. One place for the rendered references is the bottom of the README. 
* There are currently no tests, which ought to change in the near-term future.
* I have not uploaded the notebook in which I verify the function works on the `idata` Damon sent me.
* The `idata` Damon sent me _could_ exist packaged in `forecasttools-py` as an additional example `idata`.
* I need to verify whether `pandas` still needs to be in `dev` dependencies—this may be an artifact from my attempts solving the headache-inducing error from times past concerning the custom TimeStamp conversion. 

[^black]: The error received if not commented is: _Python 3.12.5 has a memory safety issue that can cause Black's AST safety checks to fail. Please upgrade to Python 3.12.6 or downgrade to Python 3.12.4_

notebooks/pyrenew_dates_to_tidy.qmd
```diff
@@ -0,0 +1,156 @@
...
+ℹ Use `spec()` to retrieve the full column specification for this data.
+ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
+Error: To use a data frame directly with `tidy_draws()`, it must already be a
+  tidy-format data frame of draws: it must have integer-like `.chain`
+  `.iteration`, and `.draw` columns with one row per draw.
+
+  The `.draw` column in the input data frame has more than one row per draw
+  (its values are not unique).
+Execution halted
+```
```
AFg6K7h4fhy2, line 156: @dylanhmorris I would appreciate if you could look at the file `idata_to_tidy.py` in `./forecasttools` and aid me with solving this error.

assets/external/idata_csv_to_tidy.R
damonbayer, line 1: No need to include if this is just repeating a function from forecasttools-R

assets/external/inference_data_tidy.R
damonbayer, line 1: No need to include if this is just repeating a function from forecasttools-R

assets/misc/references.bib
damonbayer, line 1: I don't think these references are useful, especially if they are not referred to in any documentation or readme. In the case that they were to be used in a readme or documentation, I don't see any advantage to storing a separate `.bib` file instead of just writing the reference in the text directly.

In addition, the `pyproject.toml` is a clear record of dependency between this library and jax/numpyro.

dylanhmorris: I was always skeptical here, but @AFg6K7h4fhy2 persuaded me. @AFg6K7h4fhy2 plans to write documentation/tutorials in quarto markdown. I think a `.bib` file is a reasonable (though not required) way to handle references for those docs.

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,71 @@
...
+their groups) in tidy-usable objects.
+"""
+
+import re
+
+import arviz as az
+import polars as pl
+
+
+def convert_idata_forecast_to_tidydraws(
```
damonbayer, line 14: 1. This function does not only work on forecasts, so no need to mention that in the name.
2. ArviZ uses `inference_data` in function name for [functions that work with inference data](https://python.arviz.org/en/latest/api/generated/arviz.convert_to_inference_data.html). I suggest we follow this pattern.

Thus `convert_inference_data_to_tidydraws` would be my suggestion for the name.

dylanhmorris: ^ Endorse

assets/misc/references.bib
AFg6K7h4fhy2: These are being moved to my sandbox.

forecasttools/idata_to_tidy.py
AFg6K7h4fhy2: I will follow this standard.

assets/external/inference_data_1.nc
damonbayer, line 1: Generally skeptical of putting binary files in git repos. Would prefer a simple/fast way to recreate this or a similar file that can be used for the same purposes.

AFg6K7h4fhy2: I will try to think of an alternative.

One possibility include demo-usage and then removal from the repository.

I assume you are talking about the `pyrenew` `idata`. 

The other `parquet` files and `.nc` files are package in `forecasttools` and cannot be removed without breaking the functionality of `forecasttools`.

damonbayer: Low priority, but we should think about replacing those files with scripts that generate those files (or minimal versions of those files, if they are complex to generate).

damonbayer: For example, in PyRenew-HEW we have https://github.com/CDCgov/pyrenew-hew/blob/main/pipelines/generate_test_data.R to generate "raw" data which is used in our tests, rather than storing the "raw" data itself.

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,71 @@
...
+    idata: az.InferenceData,
+    groups: list[str]
+) -> dict[str, pl.DataFrame]:
+    """
+    Creates a dictionary of polars dataframes
+    from the groups of an arviz InferenceData
+    object that when converted to a csv(s)
+    and read in R is tidy-usable.
```
damonbayer, line 22: No need to mention csv. It could be saved in any format that supports tabular data. Instead of tidy-usable, refer to the tidybayes API https://mjskay.github.io/tidybayes/articles/tidybayes.html

assets/external/inference_data_1.nc
AFg6K7h4fhy2: Seems reasonable for the `idata`, though the generation script would be long, and (not too meaningful a remark) re-introducing this code would be reversing an earlier-made decision to not have the generation code present. Thoughts for the packaged datasets used in a fair number of examples and also something that seems valuable to have a in a "forecast tools" package?

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,71 @@
...
+    Parameters
+    ----------
+    idata : az.InferenceData
+        An InferenceData object generated
+        from a numpyro forecast. Typically
+        has the groups observed_data and
+        posterior_predictive.
+    groups : list[str]
+        A list of groups belonging to the
+        idata object.
```
damonbayer, line 33: Why can't this be discerned from the idata object itself?

AFg6K7h4fhy2: Since presumably we are not interested in every group that the `idata` contains. Further thoughts?

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Returns
+    -------
+    dict[str, pl.DataFrame]
+        A dictionary of groups from the idata
+        convert to tidy-usable polars dataframe.
+    """
+    tidy_dfs = {}
+    idata_df = idata.to_dataframe()
```
damonbayer, line 42: recommend immediately converting to polars, so we aren't mixing polars and pandas throughout.

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,71 @@
...
+            if isinstance(col, tuple) and col[0] == group
+        ]
+        meta_columns = ["chain", "draw"]
+        group_df = idata_df[meta_columns + group_columns]
+        group_df.columns = [
+            col[1] if isinstance(col, tuple) else col
+            for col in group_df.columns
+        ]
+        group_pols_df = pl.from_pandas(group_df)
+        value_columns = [col for col in group_pols_df.columns if col not in meta_columns]
```
damonbayer, line 55: Use polars selectors instead https://docs.pola.rs/api/python/stable/reference/selectors.html

forecasttools/idata_to_tidy.py
AFg6K7h4fhy2: Will do, I agree with that assessment.

forecasttools/idata_to_tidy.py
dylanhmorris: Default to all groups, allow the user to specify a desired subset?

dylanhmorris: Why wouldn't we want all groups by default?

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,71 @@
...
+        group_pols_df = group_pols_df.melt(
+            id_vars=meta_columns,
+            value_vars=value_columns,
+            variable_name="variable",
+            value_name="value"
+        )
+        group_pols_df = group_pols_df.with_columns(
+            pl.col("variable").map_elements(
+                lambda x: re.sub(r"\[.*\]", "", x)).alias("variable")
```
damonbayer, line 64: Look into polars string capabilities instead? I think [`str.replace`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.replace.html) is the relevant bit.

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+        group_pols_df = group_pols_df.with_columns(
```
damonbayer, line 66: Use method chaining, rather than continually overwriting `group_pols_df`

forecasttools/idata_to_tidy.py
AFg6K7h4fhy2: > Why wouldn't we want all groups by default?

Do we want `log_likelihood` or `sample_stats` being converted / returned by default?

It seems that there would most often only be a subset of the `idata` being used.

If you want all groups, I would prefer:

> Default to all groups, allow the user to specify a desired subset?

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Returns
+    -------
+    dict[str, pl.DataFrame]
+        A dictionary of groups from the idata
+        convert to tidy-usable polars dataframe.
+    """
+    tidy_dfs = {}
+    idata_df = idata.to_dataframe()
+    for group in groups:
```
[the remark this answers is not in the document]
AFg6K7h4fhy2: Probably can, will attempt and remark if some question or issue.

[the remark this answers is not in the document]
damonbayer: Or maybe a dictionary comprehension?

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,71 @@
...
+        group_columns = [
+            col for col in idata_df.columns
+            if isinstance(col, tuple) and col[0] == group
+        ]
+        meta_columns = ["chain", "draw"]
+        group_df = idata_df[meta_columns + group_columns]
+        group_df.columns = [
+            col[1] if isinstance(col, tuple) else col
+            for col in group_df.columns
```
damonbayer, line 52: Just curious about what kinds of columns necessitate the conditional.

forecasttools/idata_to_tidy.py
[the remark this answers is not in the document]
AFg6K7h4fhy2: > Or maybe a dictionary comprehension?

This is what I was going to try.

forecasttools/idata_to_tidy.py
damonbayer: Agree we should default to all groups. We should be grabbing the list of groups from the idata object anyway, so we can check that the group is valid before trying to pull it.

damonbayer (changes requested):
Did an initial review for style / code smells. Have not tested for correctness yet.

AFg6K7h4fhy2:
> Did an initial review for style / code smells. Have not tested for correctness yet.

Thank you for this [the thank you is usually implicit, but I feel that there were enough remarks in this first review to warrant writing this message].

notebooks/pyrenew_dates_to_tidy.qmd
dylanhmorris: What does the actual tibble look like? Which draws are repeated? Is it the first one that errors?

dylanhmorris: IOW best place to start is by inspecting the output.

forecasttools/idata_to_tidy.py
```diff
@@ -29,42 +28,38 @@ def convert_inference_data_to_tidydraws(
         posterior_predictive.
     groups : list[str]
         A list of groups belonging to the
-        idata object.
+        idata object. Defaults to all groups
+        in the InferenceData.
```
dylanhmorris, line 32: This documentation should say something like "A list of groups to transform to tidy draws format" or similar.

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,67 @@
...
+                variable_name="variable",
+                value_name="value"
+            ).with_columns(
+                pl.col("variable").str.replace(r"\[.*\]", "").alias("variable")
+            ).with_columns(
+                ((pl.col("draw") - 1) % pl.col("draw").n_unique() + 1).alias(".iteration")
+            ).rename({"chain": ".chain", "draw": ".draw"}).select([".chain", ".iteration", ".draw", "variable", "value"])
+        )
+        for group in groups
+    }
```
damonbayer, line 66: Line lengths are very long. Do you have a formatter in this project? 

In particular, I think readability would be enhanced by splitting something like
```python
group_cols = [col for col in idata_df.columns if col.startswith(f"('{group}',")]
renamed_cols = {col: col.split(", ")[1].strip("')") for col in group_cols}
```

notebooks/pyrenew_dates_to_tidy.qmd
damonbayer: Is this bug still present?

damonbayer (commented):
Reviewed again for code style. Can review for correctness if known bugs are squashed.

forecasttools/idata_to_tidy.py
damonbayer: This is also a place where you can use [polars selectors](https://docs.pola.rs/api/python/stable/reference/selectors.html), rather than grabbing `idata_df.columns` explicitly.

damonbayer: If I'm not mistaken, some of the lines here appear to be more than 100 characters long.

damonbayer: Using polars selectors seems like a better option than splitting the lines up.

notebooks/pyrenew_dates_to_tidy.qmd
dylanhmorris: Also wondering this @AFg6K7h4fhy2

forecasttools/idata_to_tidy.py
dylanhmorris: Agree with @damonbayer on all counts.

AFg6K7h4fhy2: Some failed examples for posterior:

<details markdown=1>
```python
```{python}
import polars.selectors as cs
import polars as pl


df = pl.DataFrame(
    {
        "chain": [0, 1, 2, 3],
        "draw": [10, 20, 30, 40],
        "(a, var1)": [0.1, 0.2, 0.3, 0.4],
        "(a, var2)": [1.1, 1.2, 1.3, 1.4],
        "(b, var3)": [2.1, 2.2, 2.3, 2.4] 
    }
)

# want to use something like
# pl.col("*").filter(pl.col().name.str.starts_with(f"('{group}',"))

tidy_dfs = {
    group: df.select(
        #pl.col("chain", "draw"), 
        "chain",
        "draw",
        #pl.col(f"^\\({group},")
        #pl.col([col for col in df.columns if col.startswith(f"({group},")])
        #pl.col(f"^\({group},")
        #pl.col(f"^\\({group},")
        cs.starts_with(f"({group},")
        #pl.col("*").name.str.starts_with(f"('{group}',")
        #pl.col(pl.col().name.str.starts_with(f"('{group}',"))
    )
    for group in ["a", "b"]
}

print(tidy_dfs)
```
</details>

AFg6K7h4fhy2:
Requested review from @dylanhmorris. Once a review is received, the author will request another review from Damon.

.pre-commit-config.yaml
```diff
@@ -69,42 +68,72 @@ repos:
...
+    # designed as a drop-in replacement for Flake8 (plus dozens of plugins),
+    # isort, pydocstyle, pyupgrade, autoflake, and more
+    -   id: ruff
+        args: [
+          # ambiguous variable name: {name}
+          "--ignore=E741",
+          # do not assign a lambda expression, use a def
+          "--ignore=E731",
+          # indentation contains tabs (sorry, I like tabs)
+          "--ignore=W191",
```
dylanhmorris, line 94: [PEP8 strongly recommends spaces over tabs](https://peps.python.org/pep-0008/#tabs-or-spaces) for indenting Python code. You can configure your editor to insert four (or however many) spaces into Python code when you hit the tab key.

forecasttools/__init__.py
dylanhmorris, line 1: Still not convinced we should have such a large top-level namespace, but that's not for this PR

forecasttools/daily_to_epiweekly.py
```diff
@@ -8,7 +8,7 @@
 import epiweeks
 import polars as pl
 
-from forecasttools.utils import ensure_listlike
+import forecasttools
```
dylanhmorris, line 11: This risks creating a recursive import. If it doesn't currently that's only because `daily_to_epiweekly`'s functions are some of the only ones that aren't exposed by the top-level module. Should revert.

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+    Creates a dictionary of polars dataframes
+    from the groups of an arviz InferenceData
+    object for use with the tidybayes API.
+
+    Parameters
+    ----------
+    idata : az.InferenceData
+        An InferenceData object generated
+        from a numpyro forecast. Typically
```
dylanhmorris, line 25: Doesn't have to be numpyro or from a forecast. No need to name the groups.

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,99 @@
...
+    available_groups = list(idata.groups())
+    if groups is None:
+        groups = available_groups
+    else:
+        invalid_groups = [
+            group for group in groups if group not in available_groups
+        ]
+        if invalid_groups:
+            raise ValueError(
+                f"Invalid groups provided: {invalid_groups}."
```
dylanhmorris, line 48:
```suggestion
                f"Requested groups {invalid_groups} not found in this InferenceData object."
```

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,99 @@
...
+                id_vars=[".chain", ".iteration"],
+                variable_name="variable",
+                value_name="value",
+            )
+            .with_columns(
+                pl.col("variable").str.replace(r"\[.*\]", "").alias("variable")
+            )
+            .with_columns(
+                (pl.col(".iteration") + 1).alias(".iteration"),
+                (pl.col(".chain") + 1).alias(".chain"),
```
dylanhmorris, line 76: Don't need the `.alias` for modify in place.
```suggestion
                pl.col(".iteration") + 1,
                pl.col(".chain") + 1
```

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,99 @@
...
+            )
+            .with_columns(
+                (pl.col(".iteration").n_unique()).alias("draws_per_chain"),
+            )
+            .with_columns(
+                (
+                    ((pl.col(".chain") - 1) * pl.col("draws_per_chain"))
+                    + pl.col(".iteration")
+                ).alias(".draw")
+            )
```
dylanhmorris, line 86: If you do this earlier you don't need the -1.

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,99 @@
...
+            .rename(
+                {
+                    col: col.split(", ")[1].strip("')")
+                    for col in idata_df.columns
+                    if col.startswith(f"('{group}',")
+                }
+            )
+            # draw in arviz is iteration in tidybayes
+            .rename({"draw": ".iteration", "chain": ".chain"})
+            .melt(
```
dylanhmorris, line 66: `unpivot` is now preferred to `melt`
https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.unpivot.html#polars.DataFrame.unpivot

forecasttools/idata_to_tidy.py
dylanhmorris, line 86: Need to drop `"draws_per_chain"`, but also it's not a given that all chains will have the same number of draws. Instead, more robust do compute this as `.iteration + <n_draws_in_all_previous_chains`. Many ways to do that in polars.

forecasttools/idata_w_dates_to_df.py
```diff
@@ -175,7 +175,7 @@ def add_time_coords_to_idata_dimension(
     )
     # will change when xarray supports
     # numpy.dtype("datetime64[D]")
-    if not interval_dates.dtype == np.dtype("datetime64[ns]"):
+    if interval_dates.dtype != np.dtype("datetime64[ns]"):
```
dylanhmorris, line 178: neither of these is totally robust. See https://stackoverflow.com/questions/26921836/correct-way-to-test-for-numpy-dtype

Suggestion: use `issubdtype`. https://numpy.org/doc/stable/reference/generated/numpy.issubdtype.html

dylanhmorris: Out of scope for this PR, but please create an issue.

tests/test_idata_to_tidy.py
```diff
@@ -0,0 +1,53 @@
...
+import polars as pl
+import pytest
+import xarray as xr
+
+import forecasttools
+
+
+@pytest.fixture
+def mock_inference_data():
+    np.random.seed(42)
```
dylanhmorris, line 12: Running `np.random.seed` changes global state. Better practice to do something like this https://builtin.com/data-science/numpy-random-seed

tests/test_idata_to_tidy.py
```diff
@@ -0,0 +1,53 @@
...
+    posterior_predictive = xr.Dataset(
+        {
+            "observed_hospital_admissions": ("chain", np.random.randn(2, 100)),
+        },
+        coords={"chain": [0, 1]},
+    )
```
dylanhmorris, line 18: Why not run the test on the provided `inference_data_1.nc`? Or are you planning to remove it?

tests/test_idata_to_tidy.py
```diff
@@ -0,0 +1,53 @@
...
+    assert isinstance(result, dict)
+    assert "posterior_predictive" in result
+    assert isinstance(result["posterior_predictive"], pl.DataFrame)
+
+    df = result["posterior_predictive"]
+    assert all(
+        col in df.columns
+        for col in [".chain", ".draw", ".iteration", "variable", "value"]
+    )
+
```
dylanhmorris, line 38: Would be good to check that individual values are as expected, not just that the draws are unique and one for each row.

dylanhmorris (changes requested):
Thanks @AFg6K7h4fhy2. Getting close but needs a few tweaks.

tests/test_idata_to_tidy.py
AFg6K7h4fhy2: I am on the fence about removing it. Seems good to have a canonical `pyrenew-hew` `.nc` file on hand esp. given that `forecasttools-py` does / will even more so interface abundantly with `pyrenew` models. On the other hand, having adequate and general `idata` / `xarray` representations seems good for testing too. I do not know if the latter must exist at the cost of the former. I lean towards having both, with the `.nc` file perhaps being used in notebooks and the "fake" `idata`s being used for testing.

forecasttools/__init__.py
AFg6K7h4fhy2: Issue made here #62 .

.pre-commit-config.yaml
damonbayer, line 1: Suggest doing this as a separate PR.

AFg6K7h4fhy2:
A small note that the following historically worked but assumes all chains have the same number of iterations (in tidy-lingo):

<details markdown=1>

```python
tidy_dfs = {
        group: (
            idata_df.select("chain", "draw", cs.starts_with(f"('{group}',"))
            .rename(
                {
                    col: col.split(", ")[1].strip("')")
                    for col in idata_df.columns
                    if col.startswith(f"('{group}',")
                }
            )
            # draw in arviz is iteration in tidybayes
            .rename({"draw": ".iteration", "chain": ".chain"})
            .unpivot(
                index=[".chain", ".iteration"],
                variable_name="variable",
                value_name="value",
            )
            .with_columns(
                pl.col("variable").str.replace(r"\[.*\]", "").alias("variable")
            )
            .with_columns(pl.col(".iteration") + 1, pl.col(".chain") + 1)
            .with_columns(
                (pl.col(".iteration").n_unique()).alias("draws_per_chain"),
            )
            .with_columns(
                (
                    ((pl.col(".chain") - 1) * pl.col("draws_per_chain"))
                    + pl.col(".iteration")
                ).alias(".draw")
            )
            .pivot(
                values="value",
                index=[".chain", ".iteration", ".draw"],
                columns="variable",
                aggregate_function="first",
            )
            .sort([".chain", ".iteration", ".draw"])
        )
        for group in groups
    }
```

</details>

The method which does take into account the number of iterations per chain:

<details markdown=1>

```python
tidy_dfs = {
        group: (
            idata_df.select("chain", "draw", cs.starts_with(f"('{group}',"))
            .rename(
                {
                    col: col.split(", ")[1].strip("')")
                    for col in idata_df.columns
                    if col.startswith(f"('{group}',")
                }
            )
            # draw in arviz is iteration in tidybayes
            .rename({"draw": ".iteration", "chain": ".chain"})
            .unpivot(
                index=[".chain", ".iteration"],
                variable_name="variable",
                value_name="value",
            )
            .with_columns(
                pl.col("variable").str.replace(r"\[.*\]", "").alias("variable")
            )
            .with_columns(
                pl.col(".iteration") + 1, 
                pl.col(".chain") + 1)
            .pivot(
                values="value",
                index=[".chain", ".iteration"],
                columns="variable",
                aggregate_function="first",
            )
            .sort([".chain", ".iteration", ".draw"])
            .with_row_count(name=".draw", offset=1)
        )
        for group in groups
    }
```

</details>

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,84 @@
...
+their groups) in tidy-usable objects.
+"""
+
+import arviz as az
+import polars as pl
+import polars.selectors as cs
+
+
+def convert_inference_data_to_tidydraws(
+    idata: az.InferenceData, groups: list[str]
```
dylanhmorris, line 14:
```suggestion
    idata: az.InferenceData, groups: list[str] = None
```

forecasttools/idata_to_tidy.py
```diff
@@ -0,0 +1,84 @@
...
+            )
+            .with_columns(
+                pl.col("variable").str.replace(r"\[.*\]", "").alias("variable")
+            )
+            .with_columns(pl.col(".iteration") + 1, pl.col(".chain") + 1)
+            .pivot(
+                values="value",
+                index=[".chain", ".iteration"],
+                columns="variable",
+                aggregate_function="first",
```
dylanhmorris, line 77: should be `None` [per the docs](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.pivot.html), no? (unless I'm misunderstanding what your goal is with this operation). You want one column for each unique value of `"variable"` for a given `".chain"` and `".iteration"`.

```suggestion
                aggregate_function=None,
```

https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.pivot.html

tests/test_idata_to_tidy.py
```diff
@@ -0,0 +1,173 @@
...
+        A fixture providing a test InferenceData-like
+        object with two groups.
+    """
+    result = forecasttools.convert_inference_data_to_tidydraws(
+        simple_inference_data, groups=None
+    )
+    assert set(result.keys()) == set(simple_inference_data.groups())
+
+
+def test_tidydraws_format(simple_inference_data):
```
dylanhmorris, line 127: This is nice, but I think it's important also to check the correct tidying of array-valued parameters. The example idata and associated test from forecasttools-R are nice and should be easily portable to Python

https://github.com/CDCgov/forecasttools/blob/main/data-raw/ex_inferencedata_dataframe.R
https://github.com/CDCgov/forecasttools/blob/main/tests/testthat/test_inferencedata_dataframe_to_tidydraws.R

dylanhmorris (changes requested):
A few small but important things. Thanks, @AFg6K7h4fhy2!

forecasttools/idata_to_tidy.py
AFg6K7h4fhy2: For the actual `pyrenew` inference data object, but not for the other examples I have made, using `None` and not `first` produces this error:

> FAILED tests/test_idata_to_tidy.py::test_posterior_predictive_group - polars.exceptions.ComputeError: found multiple elements in the same group, please specify an aggregation function

tests/test_idata_to_tidy.py
AFg6K7h4fhy2: I will save this task for 2025-03-14.

forecasttools/idata_to_tidy.py
dylanhmorris: That suggests something is not correct upstream. Perhaps the variable name regex?

AFg6K7h4fhy2: I believe the source of this issue is the line in `col: col.split(", ")[1].strip("')")`

```
.rename(
    {
        col: col.split(", ")[1].strip("')")
        for col in idata_df.columns
        if col.startswith(f"('{group}',")
    }
)
```

AFg6K7h4fhy2: I changed

`col: col.split(", ")[1].strip("')")`

to

`col: re.search(r",\s*'?(.+?)'?\)", col).group(1)`

which I got incorrect a handful of times but which I believe works now. 

The target of this regex looks like, e.g.,  `"('posterior', 'alpha')"`. 

In the expression, the `,` matches comma; in `\s*`, the `\s` is spaces and `*` for zero or more spaces; in `'?`, the `'` means single quote and `?` means optional (0 or 1 times); in `(.+?)`, the `(` and `)` capture whatever is in the parenthesis with `.` means non-newline character, `+` means one or more characters, and `?` (which I am not sure is necessary to include here) tries to get this capture in as small a job as possible (this capture will get `alpha` in the example target); the `'?` is the same as before; the `\)` is the single closing parenthesis after the group (the `\` is needed for escaping parenthesis; and `.group(1)` in `re` gets the first item captured by `(.+?)`.

forecasttools/idata_to_tidy.py
```diff
@@ -54,7 +56,7 @@ def convert_inference_data_to_tidydraws(
             idata_df.select("chain", "draw", cs.starts_with(f"('{group}',"))
             .rename(
                 {
-                    col: col.split(", ")[1].strip("')")
+                    col: re.search(r",\s*'?(.+?)'?\)", col).group(1)
```
dylanhmorris, line 59: Feels like this would be better handled by providing a lambda rename mapping to `.rename()` rather than via dictionary comprehension. See https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.rename.html

AFg6K7h4fhy2: Hadn't thought of this. In the back of mind there are reservations I've seen from others that I haven't verified regarding about lambda mappings but they're not sufficient to not act here. Polars `rename` seems apt. Thank you for the quick comment.

AFg6K7h4fhy2: I changed the code to:

```
.rename(
    lambda col: re.search(r",\s*'?(.+?)'?\)", col).group(1)
    if col.startswith(f"('{group}',")
    else col
)
```

got 

```
forecasttools/idata_to_tidy.py:66:40: B023 Function definition does not bind loop variable `group`
```

then changed to:


```
.rename(
    lambda col, group=group: re.search(
        r",\s*'?(.+?)'?\)", col
    ).group(1)
    if col.startswith(f"('{group}',")
    else col
)
```

which works w/ tests & linting. 

Commenting above for my future self.

dylanhmorris: 1. Readability concern: at least two different senses of "group" present here.

2. I'd make it a one argument `lambda` and exclude it from linting. Making "group" an argument that defaults to the value of `group` satisfies the linter but hurts readability imo 
```
forecasttools/idata_to_tidy.py:66:40: B023 Function definition does not bind loop variable `group`
```

forecasttools/idata_to_tidy.py
damonbayer: Can you write a test to demonstrate the case where the previous version was failing and the new version succeeds? They both work correctly on the example provided in your comment: `"('posterior', 'alpha')"`.

AFg6K7h4fhy2: The split approach did not work on the InferenceData object from `pyrenew-hew` in tests when the aggregate function was not set to `first`. Yes, both the split and `re` approaches work on "('posterior', 'alpha')".

AFg6K7h4fhy2: Thank you for comment + I will write test.

forecasttools/idata_to_tidy.py
AFg6K7h4fhy2: Re: <https://github.com/CDCgov/forecasttools-py/pull/36#discussion_r2001563816>

Will do. I agree, `group=group` seems extraneous but my usual response is to defer to the linter.

forecasttools/idata_to_tidy.py
AFg6K7h4fhy2: > The split approach did not work on the InferenceData object from pyrenew-hew in tests when the aggregate function was set to None but did work when the aggregate function was set to first. Yes, both the split and re approaches work on "('posterior', 'alpha')".

I should probably investigate why, exactly, aggregate=None w/ split doesn't work for the Pyrenew inference data.

Issue #45, opened 2024-11-25
Change deprecated `importlib.resources.path` call


dylanhmorris:
```forecasttools/__init__.py:31
  /home/runner/work/forecasttools-py/forecasttools-py/forecasttools/__init__.py:31: DeprecationWarning: path is deprecated. Use files() instead. Refer to https://importlib-resources.readthedocs.io/en/latest/using.html#migrating-from-legacy for migration advice.
    with importlib.resources.path(
```

AFg6K7h4fhy2:
Worth noting that this issue is also related to #42 ; the two can likely be covered in the same PR.

AFg6K7h4fhy2:
Closed via #48 .

Issue #42, opened 2024-11-19
An Anomaly With NetCDF Bundling


O957:
This issue describes a previously undetected error that arrives with the author's [updating](https://github.com/CDCgov/forecasttools-py/pull/27/files#diff-81eac04473fd8ae59afb7b15af99d1086bd59a1e7667683b4d71760dbe0245da) of `__init__.py`.

Presently, the following exists (the author believes correctly) in `pyproject.toml`: 

```
packages = [{include = "forecasttools"}]
include = [
    { path = "forecasttools/location_table.parquet", format = "sdist" },
    { path = "forecasttools/location_table.parquet", format = "wheel" },
    { path = "forecasttools/example_flusight_submission.parquet", format = "sdist" },
    { path = "forecasttools/example_flusight_submission.parquet", format = "wheel" },
    { path = "forecasttools/example_flu_forecast_wo_dates.nc", format = "sdist" },
    { path = "forecasttools/example_flu_forecast_wo_dates.nc", format = "wheel" },
    { path = "forecasttools/example_flu_forecast_w_dates.nc", format = "sdist" },
    { path = "forecasttools/example_flu_forecast_w_dates.nc", format = "wheel" },
    { path = "forecasttools/nhsn_hosp_COVID.parquet", format = "sdist" },
    { path = "forecasttools/nhsn_hosp_COVID.parquet", format = "wheel" },
    { path = "forecasttools/nhsn_hosp_flu.parquet", format = "sdist" },
    { path = "forecasttools/nhsn_hosp_flu.parquet", format = "wheel" },
]
```

However the following:

```python
# location table (from Census data)
with importlib.resources.files(__package__).joinpath(
    "location_table.parquet"
).open("rb") as f:
    location_table = pl.read_parquet(f)

# load example flusight submission
with importlib.resources.files(__package__).joinpath(
    "example_flusight_submission.parquet"
).open("rb") as f:
    example_flusight_submission = pl.read_parquet(f)

# load example fitting data for COVID (NHSN, as of 2024-09-26)
with importlib.resources.files(__package__).joinpath(
    "nhsn_hosp_COVID.parquet"
).open("rb") as f:
    nhsn_hosp_COVID = pl.read_parquet(f)

# load example fitting data for influenza (NHSN, as of 2024-09-26)
with importlib.resources.files(__package__).joinpath(
    "nhsn_hosp_flu.parquet"
).open("rb") as f:
[...]

PR #48, opened 2024-11-25
An Anomaly With NetCDF Bundling From `importlib` Usage

AFg6K7h4fhy2:
For the scope of this PR, see issues #42 and #45 .

This PR will close both issues when merged.

### DIFF
--- a/forecasttools/__init__.py
+++ b/forecasttools/__init__.py
@@ -31,48 +31,42 @@ from forecasttools.utils import (
 location_table_path = importlib.resources.files(__package__).joinpath(
     "location_table.parquet"
 )
-location_table = pl.read_parquet(str(location_table_path))
+location_table = pl.read_parquet(location_table_path)
 
 # load example flusight submission
 example_flusight_submission_path = importlib.resources.files(
     __package__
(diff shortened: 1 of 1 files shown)


forecasttools/__init__.py
```diff
@@ -28,41 +28,52 @@
...
-    __package__, "example_flusight_submission.parquet"
-) as data_path:
-    dtypes_d = {"location": pl.Utf8}
-    example_flusight_submission = pl.read_parquet(data_path)
+example_flusight_submission_path = importlib.resources.files(
+    __package__
+).joinpath("example_flusight_submission.parquet")
+dtypes_d = {"location": pl.Utf8}
+example_flusight_submission = pl.read_parquet(
+    str(example_flusight_submission_path)
```
dylanhmorris, line 42: Is the casting to `str` really necessary?

dylanhmorris (approved):
This LGTM, but note one question

forecasttools/__init__.py
AFg6K7h4fhy2: Will verify. If not necessary, I will remove. Otherwise, keep.

Issue #19, opened 2024-10-15
Action To Use Quarto For README Generation


AFg6K7h4fhy2:
See [point 1](https://github.com/CDCgov/forecasttools-py/pull/10#discussion_r1801779639) and [point 2](https://github.com/CDCgov/forecasttools-py/pull/10#discussion_r1801781757) for additional context.

AFg6K7h4fhy2:
Moving milestone of this issue from [October 28, November 8] &rarr; [November 11, November 22] since the other PRs and issues [October 28, November 8] will quite certainty take up all the remaining time in the Sprint.

AFg6K7h4fhy2:
This issue also covers refactoring the README.

PR #49, opened 2024-11-25
Pre-Commit Hook To Use Quarto For README Generation

AFg6K7h4fhy2:
For the scope of this PR, refer to issue #19 .


AFg6K7h4fhy2:
@dylanhmorris I would appreciate some guidance on the correct format of the following workflow:

* <https://github.com/CDCgov/forecasttools-py/pull/49/files#diff-e4a6b66cdb8f46bc35fa2d206adad3e1241f2064aacf427f28094b0ec892158d>

dylanhmorris:
If you're going to have this on every push that modifies `README.qmd`, regardless of branch, should it definitely be a github action and not a pre-commit hook?

AFg6K7h4fhy2:
Re: <https://github.com/CDCgov/forecasttools-py/pull/49#issuecomment-2501521003>

I am not sure here and will defer to whatever is decided by you and those who are tagged.

Since I do not expect the README.qmd to be updated very often and since I have a vague notion of "server side rendering" is "more robust" I opted for GitHub Actions. That this could be a `pre-commit` hook did occur to me, but the aforementioned and the belief that I'd be able to implement an action quicker steered me away from this path.

dylanhmorris:
The downside of server-side is that it adds additional commits, and means that the local repo doesn't necessarily reflect what's being pushed (without an additional pull). Given that any reasonable dev environment for editing the readme _should_ be able to render it, I don't really see server side rendering as "more robust" in this case.

dylanhmorris:
> The downside of server-side is that it adds additional commits, and means that the local repo doesn't necessarily reflect what's being pushed (without an additional pull). Given that any reasonable dev environment for editing the readme _should_ be able to render it, I don't really see server side rendering as "more robust" in this case.

The one thing that will be less "just works" with client-side rendering is editing the `README.qmd` in the web gui. But that arguably should not be done anyway.

AFg6K7h4fhy2:
The additional commits & `pull` are arguments enough for me. I will wait for another hour or so for comments, working on other entities in `forecasttools-py`, before proceeding with a shift to `pre-commit`.

AFg6K7h4fhy2:
Order of operations:

* Updated `pre-commit.yaml` in `./.github/workflows/` to have more recent versions.
* Updated `.pre-commit-config.yaml` with sections.
* Ran `pre-commit autoupdate`.
* Create `./scripts/` and `render-quarto-readme.sh` within.
* Ran `chmod +x render-quarto-readme.sh` in `./scripts/hooks/` after creation.
* Update `.pre-commit-config.yaml` to have local hook.

scripts/hooks/render-quarto-readme.sh
```diff
@@ -21,5 +21,16 @@ poetry install --with dev
...
-# # add the generated README.md to the commit
-# git add README.md
+# prompt user to stage/commit README
+# if there is a change
+if [[ -f README.md ]]; then
+  if ! git diff --quiet README.md; then
+    echo "README.md has been modified. Please stage and commit."
+    exit 1
+  else
+    echo "README.md is up to date."
```
dylanhmorris, line 32: Should `exit 0` here, no?
```suggestion
    echo "README.md is up to date."
    exit 0
```

AFg6K7h4fhy2:
Differences in rendering between computer architectures (Windows, Ubuntu, etc...) might not (will likely not) yield the exact same output, causing a failure. One solution: rendering environment much be synched between remote CI and local.

dylanhmorris (approved):
LGTM

.pre-commit-config.yaml
```diff
@@ -1,43 +1,68 @@
...
+# QUARTO README RENDERING
+################################################################################
+-   repo: local
+    hooks:
+    -   id: render-quarto-readme
+        name: "render README.qmd as README.md"
+        entry: "./scripts/hooks/render-quarto-readme.sh"
+        language: script
+        always_run: true
+        files: "README.qmd"
```
dylanhmorris, line 26:
```suggestion
```

PR #50, opened 2024-11-26
Patch to 19-action-to-use-quarto

gvegayon:
Hey @AFg6K7h4fhy2, I think this fixes the last bit you need for your PR to work properly.


AFg6K7h4fhy2:
This PR will be removed once #49 (where most development has occurred) is merged. Thank you again George for your help. This small change to the PR (in addition to adding quotes in one of the lines) is what got the workflow working.

Issue #74, opened 2025-03-24
Migrate From Poetry To UV


AFg6K7h4fhy2:
This issue covers migrating from `poetry` to `uv`.

For context, this transition was decided across CFA Predict and captured recently in the guidance provided by the Predict handbook.

A similar issue can be found [here](https://github.com/CDCgov/PyRenew/issues/494).

PR #77, opened 2025-05-29
Pre-Commit Configuration Updated

O957:
This PR adds the following:

* [x] Default specifications for `pre-commit` CI (see [this comment](https://github.com/CDCgov/hubverse-annotator/pull/26#issuecomment-2919753492) regarding utility).
* [x] The `uv-lock` `pre-commit` hook.

This PR removes the following:

* [x] Duplicate "general" `pre-commit` hook specifications.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -109,7 +109,7 @@ repos:
           "--ignore=E741",
           # do not assign a lambda expression, use a def
           "--ignore=E731",
-          # indentation contains tabs (sorry, I like tabs)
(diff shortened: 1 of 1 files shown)


AFg6K7h4fhy2:
This PR is gated by #74 , since we want the `uv-lock` `pre-commit` hook, but do not yet use `uv` for this project.

damonbayer (approved):
I'm still not a fan of having non-standard pre-commit (as compared to other CFA/STF repos) here, but this seems like an improvement.

Issue #104, opened 2025-08-20
add prune_bad_chains function


damonbayer:
Migrate from https://github.com/cdcent/stf-pyrenew-added-value/

PR #107, opened 2025-08-22
Add and test prune_chains_by_rel_diff

damonbayer:
Closes https://github.com/CDCgov/forecasttools-py/issues/104.

- [x] Adds a new function `prune_chains_by_rel_diff`
	- It is designed to remove chains from an InferenceData object based on `lp`, if present, or `log likelihood`.
- [x] Adds tests for `prune_chains_by_rel_diff`
- [x] Revises `notebooks/datasets_and_idata_creation.qmd` for ruff formatting and producing a `example_flu_forecast_wo_dates.nc` with `lp`.
- [x] Updates `example_flu_forecast_wo_dates.nc` to include `lp`.
- [x] adds pyarrow dependency to work around this somewhat strange error message when using `pl.from_pandas`:
```pycon
pyarrow is required for converting a pandas dataframe to Polars, unless each of its columns is a simple numpy-backed one (e.g. 'int64', 'bool', 'float32' - not 'Int64')
```


forecasttools/arviz.py
```diff
@@ -154,3 +155,76 @@ def assign_coords_from_start_step(
...
+
+    Parameters
+    ----------
+    idata : az.InferenceData
+        ArviZ InferenceData object containing MCMC samples and statistics.
+    rel_diff_thresh : float, default 0.9
+        Relative difference threshold for keeping chains. Chains with relative
+        performance above this threshold compared to the best chain are kept.
+        Value should be between 0 and 1, where 1 means only keep chains identical
+        to the best, and 0 means keep all chains.
```
dylanhmorris, line 178: Justification for the default?

dylanhmorris (approved):
A couple minor questions, but this LGTM

forecasttools/arviz.py
dylanhmorris: Yes, I think without a clear standard for the default should let the user make the call.

forecasttools/arviz.py
```diff
@@ -154,3 +155,76 @@ def assign_coords_from_start_step(
         return None
     else:
         return out
+
+
+def prune_chains_by_rel_diff(
```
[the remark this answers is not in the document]
damonbayer: It appears to be a simple rename https://github.com/arviz-devs/arviz/blob/f89cdee961471c02f1391e51e0a0c8272b2639ea/arviz/data/io_numpyro.py#L233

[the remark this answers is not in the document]
dylanhmorris: We should maybe open an issue?

forecasttools/arviz.py
damonbayer: [`367e687` (#107)](https://github.com/CDCgov/forecasttools-py/pull/107/commits/367e68719bdf573ea75375d89944d69a557d3068)

PR #222, opened 2026-07-09
daily-to-weekly aggregation utility with epiweek/isoweek support

sbidari:
What changed from previous forecasttools-py version:
- support for two standards: epiweek and isoweek
- option to append column with week start/end date (this is needed for backtest eval data)

### DIFF
--- a/cfa/stf/forecasttools/aggregate_to_weekly.py
+++ b/cfa/stf/forecasttools/aggregate_to_weekly.py
@@ -234,28 +234,32 @@ def daily_to_weekly(
     ).unnest("week_struct")
 
     group_cols = ["week", "weekyear"] + id_cols
-    n_elements = df.group_by(group_cols).agg(pl.len().alias("n_elements"))
-    problematic_trajectories = n_elements.filter(pl.col("n_elements") > 7)
-    if not problematic_trajectories.is_empty():
+    grouped_df = df.group_by(group_cols).agg(
+        pl.col(value_col).sum().alias(weekly_value_name),
+        pl.len().alias("n_dates"),
+        pl.col(date_col).dt.date().n_unique().alias("n_unique_dates"),
+    )
+
+    duplicate_date_trajectories = grouped_df.filter(
+        pl.col("n_dates") != pl.col("n_unique_dates")
+    )
+    if not duplicate_date_trajectories.is_empty():
         message = (
-            "At least one trajectory has more than 7 values for a given "
+            "At least one trajectory has repeated dates within a given "
             f"week.\n"
-            f"Problematic trajectories with more than 7 values: {problematic_trajectories}"
+            f"Trajectories with repeated dates: {duplicate_date_trajectories}"
         )
         raise ValueError(message)
 
     if strict:
-        valid_groups = n_elements.filter(pl.col("n_elements") == 7)
-        df = df.join(valid_groups.select(group_cols), on=group_cols, how="inner")
+        grouped_df = grouped_df.filter(pl.col("n_unique_dates") == 7)
 
-    df = (
-        df.group_by(group_cols)
-        .agg(pl.col(value_col).sum().alias(weekly_value_name))
-        .sort(group_cols)
+    grouped_df = grouped_df.drop(["n_dates", "n_unique_dates"]).sort(
+        ["weekyear", "week", *id_cols]
     )
 
     if with_week_start_date:
-        df = df.with_columns(
+        grouped_df = grouped_df.with_columns(
             pl.struct(["weekyear", "week"])
             .map_elements(
                 lambda elt: _calculate_week_startdate(
@@ -269,7 +273,7 @@ def daily_to_weekly(
         )
 
     if with_week_end_date:
-        df = df.with_columns(
+        grouped_df = grouped_df.with_columns(
             pl.struct(["weekyear", "week"])
             .map_elements(
                 lambda elt: _calculate_week_enddate(
@@ -282,4 +286,4 @@ def daily_to_weekly(
             .alias(week_end_date_name)
         )
 
-    return df
+    return grouped_df
--- a/cfa/stf/forecasttools/aggregate_to_weekly.py
+++ b/cfa/stf/forecasttools/aggregate_to_weekly.py
@@ -59,7 +59,7 @@ def _calculate_week_enddate(
 
 
 def floor_week(
-    date: datetime.date,
+    date: datetime.date | datetime.datetime,
     standard: WeekStandard | str = "MMWR",
 ) -> datetime.date:
     """Get the first date in an epidemiological week.
@@ -91,7 +91,7 @@ def floor_week(
 
 
 def ceiling_week(
-    date: datetime.date,
+    date: datetime.date | datetime.datetime,
     standard: WeekStandard | str = "MMWR",
 ) -> datetime.date:
     """Get the last date in an epidemiological week.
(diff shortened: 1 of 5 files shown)


dylanhmorris:
`epiweeks` also supports isoweeks, so for simplicity of implementation consider using it for both?

cfa/stf/forecasttools/aggregate_to_weekly.py
```diff
@@ -0,0 +1,196 @@
...
+
+def calculate_week_and_year(
+    date_input: datetime.date,
+    standard: WeekStandard = "epiweek",
+) -> dict[str, int]:
+    """Convert a date to week and weekyear for the requested weekly standard."""
+    if not isinstance(date_input, datetime.date):
+        raise TypeError(
+            f"date_input must be a datetime.date, got {type(date_input).__name__}"
+        )
```
damonbayer, line 31: This function should support `datetime.datetime`s as well.

damonbayer (changes requested):
Two requests:
1. Can we use the same language as in R forecasttols:: https://cdcgov.github.io/forecasttools/reference/epiweek_fn.html. e.g. `mmwr` vs `iso`.
2. Can you also add the floor and ceiling functions?

sbidari:
> 2. Can you also add the floor and ceiling functions?

These are analogues to `epiweek_start` and `epiweek_end` from R forecasttools? https://cdcgov.github.io/forecasttools/reference/epiweek_start.html

damonbayer:
> > 2. Can you also add the floor and ceiling functions?
> 
> These are analogues to `epiweek_start` and `epiweek_end` from R forecasttools? [cdcgov.github.io/forecasttools/reference/epiweek_start.html](https://cdcgov.github.io/forecasttools/reference/epiweek_start.html)

These: https://cdcgov.github.io/forecasttools/reference/floor_epiweek.html

cfa/stf/forecasttools/aggregate_to_weekly.py
sbidari: epiweeks supports `datetime.date` and `datetime.datetime`, updated type annotation and error message to make this explicit https://github.com/CDCgov/cfa-stf-forecasttools/pull/222/commits/b833f2a05c9189064cacf22f351c314ec22e442b

sbidari:
> `epiweeks` also supports isoweeks, so for simplicity of implementation consider using it for both?

done https://github.com/CDCgov/cfa-stf-forecasttools/pull/222/commits/b833f2a05c9189064cacf22f351c314ec22e442b

cfa/stf/forecasttools/aggregate_to_weekly.py
```diff
@@ -0,0 +1,285 @@
...
+        .alias("week_struct")
+    ).unnest("week_struct")
+
+    group_cols = ["week", "weekyear"] + id_cols
+    n_elements = df.group_by(group_cols).agg(pl.len().alias("n_elements"))
+    problematic_trajectories = n_elements.filter(pl.col("n_elements") > 7)
+    if not problematic_trajectories.is_empty():
+        message = (
+            "At least one trajectory has more than 7 values for a given "
+            f"week of a given weekyear.\n"
```
damonbayer:
```suggestion
            f"week.\n"
```

damonbayer (changes requested):
@sbidari some codex-identified bugs. These tests should pass.
```python
import datetime as dt

import polars as pl

from cfa.stf.forecasttools import daily_to_weekly


def test_strict_mode_accepts_duplicate_date_and_missing_day():
    """Seven rows are accepted even though they contain only six distinct dates."""
    sunday = dt.date(2026, 1, 4)
    dates = [
        sunday,
        sunday,  # Duplicate
        sunday + dt.timedelta(days=1),
        sunday + dt.timedelta(days=2),
        sunday + dt.timedelta(days=3),
        sunday + dt.timedelta(days=4),
        sunday + dt.timedelta(days=5),
        # Saturday is missing
    ]

    df = pl.DataFrame(
        {
            ".draw": [0] * 7,
            "date": dates,
            "value": [1] * 7,
        }
    )

    result = daily_to_weekly(df, strict=True)

    # Expected: the incomplete week is omitted or rejected.
    # Actual: one weekly total is returned because row count equals seven.
    assert result.is_empty()


def test_strict_mode_drops_null_identifier_group():
    """A complete trajectory is lost when its identifier is null."""
    sunday = dt.date(2026, 1, 4)

    df = pl.DataFrame(
        {
            "id": pl.Series([None] * 7, dtype=pl.String),
            "date": [sunday + dt.timedelta(days=i) for i in range(7)],
            "value": [1] * 7,
        }
    )

    result = daily_to_weekly(df, id_cols="id", strict=True)

    # Expected: one complete group with weekly_value == 7.
    # Actual: the null join key does not match, so the result is empty.
    assert result.height == 1
    assert result["weekly_value"].item() == 7


def test_multi_year_output_is_not_chronological():
    """Week number is sorted before week-year."""
    start = dt.date(2025, 12, 21)

    df = pl.DataFrame(
        {
            ".draw": [0] * 21,
            "date": [start + dt.timedelta(days=i) for i in range(21)],
            "value": [1] * 21,
        }
    )

    result = daily_to_weekly(df)

    actual = list(zip(result["weekyear"], result["week"], strict=True))

    # Expected chronological ordering.
    # Actual: [(2026, 1), (2025, 52), (2025, 53)]
    assert actual == [(2025, 52), (2025, 53), (2026, 1)]
```

cfa/stf/forecasttools/aggregate_to_weekly.py
```diff
@@ -0,0 +1,285 @@
...
+        raise ValueError(message)
+
+    if strict:
+        valid_groups = n_elements.filter(pl.col("n_elements") == 7)
+        df = df.join(valid_groups.select(group_cols), on=group_cols, how="inner")
+
+    df = (
+        df.group_by(group_cols)
+        .agg(pl.col(value_col).sum().alias(weekly_value_name))
+        .sort(group_cols)
```
damonbayer: **[P2] Sort by week-year before week**

Sorting by `week` before `weekyear` makes multi-year output nonchronological. For three consecutive MMWR weeks spanning 2025–2026, the current result is `(2026, 1), (2025, 52), (2025, 53)` instead of chronological order. Please sort each trajectory by `weekyear`, then `week`, typically with ID columns first.

— Codex

cfa/stf/forecasttools/aggregate_to_weekly.py
```diff
@@ -0,0 +1,285 @@
...
+        message = (
+            "At least one trajectory has more than 7 values for a given "
+            f"week of a given weekyear.\n"
+            f"Problematic trajectories with more than 7 values: {problematic_trajectories}"
+        )
+        raise ValueError(message)
+
+    if strict:
+        valid_groups = n_elements.filter(pl.col("n_elements") == 7)
+        df = df.join(valid_groups.select(group_cols), on=group_cols, how="inner")
```
damonbayer: **[P2] Preserve groups with null identifiers**

This inner join uses the Polars default `nulls_equal=False`. A complete seven-day trajectory whose ID is null appears in `valid_groups` but fails to join back, so strict mode silently returns no row; the same input returns a total with `strict=False`. Please use `nulls_equal=True`, explicitly reject null identifiers, or filter complete groups without a null-sensitive join.

— Codex

cfa/stf/forecasttools/aggregate_to_weekly.py
```diff
@@ -0,0 +1,285 @@
...
+            lambda elt: _calculate_week_and_year(elt, standard=standard),
+            return_dtype=pl.Struct(
+                [pl.Field("week", pl.Int64), pl.Field("weekyear", pl.Int64)]
+            ),
+        )
+        .alias("week_struct")
+    ).unnest("week_struct")
+
+    group_cols = ["week", "weekyear"] + id_cols
+    n_elements = df.group_by(group_cols).agg(pl.len().alias("n_elements"))
```
damonbayer: **[P2] Count distinct dates in strict mode**

`pl.len()` counts rows, not dates. For example, seven rows containing a duplicated Sunday, Monday–Friday, and no Saturday pass `strict=True` and produce a weekly total even though the group has only six distinct dates. Please validate uniqueness per trajectory/date and count distinct `date_col` values before accepting the week.

— Codex

cfa/stf/forecasttools/aggregate_to_weekly.py
sbidari: done https://github.com/CDCgov/cfa-stf-forecasttools/pull/222/commits/5439fe331b30eedb9ffe7416e844bb4ef355befe

cfa/stf/forecasttools/aggregate_to_weekly.py
sbidari: Removed inner join-based approach (https://github.com/CDCgov/cfa-stf-forecasttools/pull/222/commits/d4b91c6a49edde8cd366591c7d8afc80eaa870a5)

cfa/stf/forecasttools/aggregate_to_weekly.py
sbidari: https://github.com/CDCgov/cfa-stf-forecasttools/pull/222/commits/d4b91c6a49edde8cd366591c7d8afc80eaa870a5
Now raises error when row `date counts != unique dates` within the group. This does not allow duplicate date entries for a week which I think is we want. But lmk if you disagree

damonbayer (approved):
thanks @sbidari!

PR #246, opened 2026-09-02
Conversion function  from forecast-sample table to  Hubverse sample or quantile tables

cdc-mitzimorris:
Add two small, Python/polars functions that convert a six-column forecast-sample table into the initial eight-column Hubverse sample or quantile table.  The experiment runner's canonical column names are the defaults, and callers may supply different source-column names.

These functions are in file cfa/stf/forecasttools/hubverse.py, with unit tests in file tests/cfa/stf/forecasttools/test_hubverse.py.

This PR also adds directory tests/cfa/stf/forecasttools/test_data which contains the data files needed for this test.  See the README.md in that directory.   Some test data was generated using R and R would be needed to regenerate the test data but there is no runtime R dependency.

## Reviewer notes

The new public functions are:

- samples_to_hubverse()
- samples_to_hubverse_quantiles()

Both accept an exact six-column Polars input containing draw, date, location, variable, value, and resolution. Source-column names are configurable, but the output is intentionally
limited to this initial Hubverse profile:

reference_date
target
horizon
target_end_date
location
output_type
output_type_id
value

Key behaviors:

- Callers provide target mappings, location mappings, the reference date, and the horizon unit.
- Sample output always uses Hubverse’s one-based IDs, 1..N.
- source_draw_id_base declares whether incoming sample IDs start at zero or one.
- Quantiles use linear interpolation equivalent to R type 7.
- Week horizons require target dates to be exact seven-day multiples from the reference date.
- Output schemas and row ordering are deterministic.
- Validation rejects malformed schemas, null or non-finite values, incomplete sample coverage, invalid date spacing, missing mappings, mapping collisions, and duplicate output keys.

The test data includes:

- A production-shaped synthetic PyRenew artifact.
- Static sample and quantile golden outputs generated by the existing routine-forecasting R conversion.
- Provenance, hashes, and regeneration instructions in the test-data README and supporting documentation.

Routine forecasting uses epiweekly, while this API’s input contract uses weekly; the R parity test makes that normalization explicit before calling the converter. R is required only to regenerate the golden fixtures, not to run pytest.

Suggested review focus:

- Whether the public API and fixed eight-column scope are appropriate.
- Whether strict sample-coverage validation matches expected callers.
- The zero-based to one-based sample-ID conversion.
- Horizon and quantile semantics.
- Clarity of validation errors and public docstrings.


cfa/stf/forecasttools/hubverse.py
```diff
@@ -0,0 +1,649 @@
...
+        resolution: str = series.get_column("resolution").item(0)
+        _validate_date_spacing(expected_dates, resolution=resolution)
+
+
+def _normalize_sample_ids(
+    samples: pl.DataFrame,
+    *,
+    source_draw_id_base: Literal[0, 1],
+) -> pl.DataFrame:
+    """Validate source draw IDs and return one-based Hubverse sample IDs."""
```
damonbayer: hubverse sample IDs don't have to be 0-based (they don't even have to be integers)

cfa/stf/forecasttools/hubverse.py
```diff
@@ -0,0 +1,649 @@
...
+        raise TypeError("samples must be a Polars DataFrame")
+
+    source_columns = {
+        "draw": draw_col,
+        "date": date_col,
+        "location": location_col,
+        "variable": variable_col,
+        "value": value_col,
+        "resolution": resolution_col,
+    }
```
damonbayer, line 46: I wouldn't expect every source file to have all of these, but we really haven't talked about what source files look like. Maybe we should start by defining our expectations for source files?

damonbayer (changes requested):
@cdc-mitzimorris Overall I feel uncertain about the intended use of this these functions. I know the destination format is something very close to the [schema described here](https://cdc.sharepoint.com/:x:/r/teams/CenterforForecastingandOutbreakAnalytics/_layouts/15/doc2.aspx?sourcedoc=%7BAC375896-0958-46C6-9BE9-23E6099E79CC%7D&file=Database%20Schema.xlsx&action=default&mobileredirect=true&DefaultItemOpen=1), but I'm not sure where the assumptions about the input format come from.

cdc-mitzimorris:
> @cdc-mitzimorris Overall I feel uncertain about the intended use of this these functions. I know the destination format is something very close to the [schema described here](https://cdc.sharepoint.com/:x:/r/teams/CenterforForecastingandOutbreakAnalytics/_layouts/15/doc2.aspx?sourcedoc=%7BAC375896-0958-46C6-9BE9-23E6099E79CC%7D&file=Database%20Schema.xlsx&action=default&mobileredirect=true&DefaultItemOpen=1), but I'm not sure where the assumptions about the input format come from.



The intended use is to convert the experiment scaffold’s native predictive-draw table (samples.parquet) into a Hubverse model-output table.

a) The destination is the initial eight-column Hubverse output format, not the RTM database format - that will be done in a subsequent PR. 

b) The input assumptions come from the scaffold’s current six-column predictive-draw contract, as produced by the current PyRenew models (including PyRenew HE/HEW). The source-column names are configurable for compatible outputs.

cfa/stf/forecasttools/hubverse.py
```diff
@@ -0,0 +1,649 @@
...
+            .agg(
+                pl.col("value").quantile(level, interpolation="linear").cast(pl.Float64)
+            )
+            .with_columns(
+                pl.lit("quantile").alias("output_type"),
+                pl.lit(level, dtype=pl.Float64).alias("output_type_id"),
+            )
+            for level in quantile_levels
+        ]
+    )
```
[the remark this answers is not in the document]
cdc-mitzimorris: suggested change implemented.

cfa/stf/forecasttools/hubverse.py
cdc-mitzimorris: removed this restriction.

cfa/stf/forecasttools/hubverse.py
cdc-mitzimorris: Agreed. I’ve clarified that these functions consume the experiment scaffold’s predictive-draw table, which requires draw, date, location, variable, value, and resolution. Column names remain configurable for compatible sources, and additional columns are now allowed and ignored. 
The public docstrings now document the required fields and their expected types.

damonbayer:
I think we should probably discuss these conventions a bit more face to face. I have not done a good job documenting some of the progress we have made in the rtm db project and its implications for the rest of our projects.

> a) The destination is the initial eight-column Hubverse output format, not the RTM database format - that will be done in a subsequent PR.

The rtm db is in the hubverse format, so I don’t think it makes sense to have an intermediate format. I think we should be enforcing rtm db standards in the backtesting repos.

> b) The input assumptions come from the scaffold’s current six-column predictive-draw contract, as produced by the current PyRenew models (including PyRenew HE/HEW). The source-column names are configurable for compatible outputs.

if this conversion is specific to pyrenew, then I’m not sure it belongs here under such a specific name.

cdc-mitzimorris:
> I think we should probably discuss these conventions a bit more face to face. I have not done a good job documenting some of the progress we have made in the rtm db project and its implications for the rest of our projects.

agree.   I fed ChatGPT the rtm DB excel spreadsheet as well as pointing it at the hubverse utils.   I will dig up the early parts of the conversation - the upshot was that rtm DB format still had details to be pinned down.

PR #248, opened 2026-09-08
CRPS scoring for forecast sample using scoringutils

cdc-mitzimorris:
~~Adds a small, reusable Polars API that calculates empirical continuous ranked probability scores (CRPS) from predictive samples. The implementation uses NumPy for the numerical calculation and scores each explicitly declared forecast unit independently.~~

per discussion, wrap scoringutils  CRPS calculation.

To test for parity with R scoring utils, this PR adds test fixture `tests/cfa/stf/forecasttools/test_data/scoringutils_crps/` which includes:

- a long-form input table with several forecast units;
- expected natural- and log1p-scale CRPS results;
- an R script that regenerates the expected results; and
- a short README recording the R, `scoringutils`, and `scoringRules` versions
  used to generate the fixture.

Add directory `docs` and documentation for this utility.


cfa/stf/forecasttools/scoring.py
```diff
@@ -0,0 +1,240 @@
...
+        (above_observation - probability_midpoints) * (sorted_samples - observation)
+    )
+    return float(score)
+
+
+def score_sample_crps(
+    forecasts: pl.DataFrame,
+    *,
+    forecast_unit: Sequence[str],
+    scale: Literal["natural", "log1p"],
```
damonbayer: Could/should this take a function instead of a string?

cfa/stf/forecasttools/scoring.py
```diff
@@ -0,0 +1,240 @@
...
+            f"forecast_unit must not contain reserved output column(s): {reserved}"
+        )
+
+    requested_columns = [*unit_columns, *value_columns.values()]
+    missing_columns = [
+        column for column in requested_columns if column not in forecasts.columns
+    ]
+    if missing_columns:
+        missing = ", ".join(missing_columns)
+        raise ValueError(f"forecasts is missing required column(s): {missing}")
```
damonbayer, line 88: totally unneccessary, but I like to use walrus operator for these

```python
if x := y
	raise ValueError(f"arg is bad: {x}")
```

tests/cfa/stf/forecasttools/test_data/scoringutils_crps/generate_expected.R
```diff
@@ -0,0 +1,61 @@
+expected_r_version <- "4.3.1"
+expected_scoringutils_version <- "2.1.1"
+expected_scoringrules_version <- "1.1.3"
```
damonbayer: probably drop all of these requirements? R and scoringutils are both out of date and the calculation is unlikely to change.

tests/cfa/stf/forecasttools/test_data/scoringutils_crps/generate_expected.R
```diff
@@ -0,0 +1,61 @@
...
+  units <- unique(scored_input[forecast_unit])
+  units$scale <- scale
+  units$crps <- vapply(seq_len(nrow(units)), function(index) {
+    matches <- rep(TRUE, nrow(scored_input))
+    for (column in forecast_unit) {
+      matches <- matches & scored_input[[column]] == units[[column]][index]
+    }
+    observations <- unique(scored_input$observed[matches])
+    stopifnot(length(observations) == 1L)
+    scoringutils::crps_sample(
```
damonbayer: Not sure why this script is so manual. The point of scoringutils is to make the scoring ergonomic

```r
library(scoringutils)

example_sample_continuous |>
  filter(observed >=0) |> 
  transform_forecasts(fun = log_shift, offset = 1) |> 
  score()
```

damonbayer (approved):
Thanks @cdc-mitzimorris! Some suggestions.

cfa/stf/forecasttools/scoring.py
cdc-mitzimorris: we can do this but it will require adding both the function and the label to use in the output column.

```python
score_sample_crps(
      forecasts,
      forecast_unit=forecast_unit,
      transform=np.log1p,
      scale_name="log1p",
  )
```

@damonbayer  - is this OK?  if so, will change

damonbayer: @cdc-mitzimorris LGTM

cdc-mitzimorris: implemented.

cfa/stf/forecasttools/scoring.py
cdc-mitzimorris: noted, but not changing.

tests/cfa/stf/forecasttools/test_data/scoringutils_crps/generate_expected.R
cdc-mitzimorris: dropped.

tests/cfa/stf/forecasttools/test_data/scoringutils_crps/generate_expected.R
cdc-mitzimorris: test was setup for unequal sizes of samples; but that's not the focus of the test.  simplified test for more ergonomic scoring.

